Issue #2, opened 2016-08-24
slow write BW observed beyond 64KB size


drossetti:
This has been reported by Mark Mark Silberstein mark@ee.technion.ac.il
￼
We finally pinpointed the setup, and it's easily reproducible. 
1. Get the CPU ptr for the buffer in the  mapped BAR
2. Sequentially pread from file into that buffer in blocks >=64K. 

As long as blocks are less than 64K, we get ~1GB/s. For blocks >= 64K we get around 13MB/s

drossetti:
closing as fixed

PR #3, opened 2016-08-24
detect physically contiguous ranges and call remap_pfn_ranges just once

drossetti:
fixes issue #2


drossetti:
note: this is only fixing the contiguous case. fixing the generic case needs digging in the Linux kernel x86_64 design.

Issue #7, opened 2016-09-15
add mechanism for creating RPM packages


drossetti:

drossetti:
closing

PR #8, opened 2016-09-16
add support for building RPMs

drossetti:
resolve Issue #7

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ drv_install:
 
 
 clean:
-	rm -f *.o $(EXES) lib*.{a,so} *~ core.* && \
+	rm -f *.o $(EXES) lib*.{a,so}* *~ core.* && \
 	$(MAKE) -C gdrdrv clean
 
 .PHONY: driver clean all lib exes lib_install install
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,4 +1,5 @@
 PREFIX ?= /usr/local
+DESTLIB ?= /usr/local/lib64
```
drossetti: this is not correct, if PREFIX is passed but DESTLIB is not, we want DESTLIB to change accordingly (appending /lib64 to PREFIX)

drossetti:
@spotluri there is no reliable way of expressing that requirement. E.g. when nvidia driver is installed via the .run installer, there is no init.d script associated with it.

Makefile
drossetti: fixed in latest push

init.d/gdrcopy
```diff
@@ -78,6 +91,12 @@ load_module()
 # Start daemon
 function start() {
 
+    echo -n $"Checking required modules: "
```
drossetti: @spotluri with this change, the init script will check for the presence of the nvidia driver

drossetti:
@spotluri fyi locally tested on both RH6 and 7

init.d/gdrcopy
spotluri: thanks @drossetti!

PR #17, opened 2017-06-02
add root privileges verification

giovannesant:
Insmod script uses sudo in all commands, but some machines has no sudo installed.

Removed variable THIS_DIR because this isn't used at all.


drossetti:
user ghost has been deleted
this is related to #51, so closing this one in favor of the other

drossetti:
@pakmarkthub no need to review

PR #22, opened 2017-12-06
fix memory fence bugs

### LINKED ISSUE #21: copy_to_bar: sfence not issued in the right order in some scenarios
I observed a 1msec latency from when gdr_copy_to_bar is issued to when the update is observed on the GPU.  

When the target buffer is not aligned or when the copy size is too small, gdr_copy_to_bar translates to an sfence followed by a memcpy. 

Issuing sfence after the mempcy seems to prevent some buffering and helps reduce the latency significantly.

drossetti:
a performance issue uncovered a few problems with fences:
- a few store fences are missing, 
- others are badly placed

fixes #21

### DIFF
--- a/gdrapi.c
+++ b/gdrapi.c
@@ -392,7 +392,8 @@ int gdr_copy_from_bar(void *h_ptr, const void *bar_ptr, size_t size)
         gdr_dbgc(1, "using plain implementation of gdr_copy_from_bar\n");
         memcpy(h_ptr, bar_ptr, size);
 
-        // note: fencing is not needed because plain stores are used
+        // note: fencing is not needed assuming the memcpy() implementation
+        // has all the required write fences
         //_mm_sfence();
 
     } while (0);
--- a/gdrapi.c
+++ b/gdrapi.c
@@ -350,9 +350,15 @@ int gdr_copy_to_bar(void *bar_ptr, const void *h_ptr, size_t size)
             memcpy_uncached_store_sse(bar_ptr, h_ptr, size);
             break;
         }
+
         // fall through
         gdr_dbgc(1, "using plain implementation of gdr_copy_to_bar\n");
         memcpy(bar_ptr, h_ptr, size);
+
+        // fencing is needed even for plain memcpy(), due to performance
+        // being hit by delayed flushing of WC buffers
+        _mm_sfence();
+
     } while (0);
 
     return 0;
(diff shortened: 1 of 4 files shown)


gdrapi.c
```diff
@@ -381,9 +387,14 @@ int gdr_copy_from_bar(void *h_ptr, const void *bar_ptr, size_t size)
...
             break;
         }
+
         // fall through
         gdr_dbgc(1, "using plain implementation of gdr_copy_from_bar\n");
         memcpy(h_ptr, bar_ptr, size);
+
+        // note: fencing is not needed because plain stores are used
+        //_mm_sfence();
+
```
spotluri: i am wondering if this comment is relevant give these are loads from the BAR

drossetti: my rationale is:
- reading from a BAR does not require a write fence
- writing to host memory via memcpy() does not require a write fence, as I assume that even if memcpy() were implemented using NT stores ( compiler / libc detail ) a correct implementation would also include a trailing write fence
so the comment should really be:
"note: assuming the memcpy() implementation has all the required write fences

memcpy_sse41.c
```diff
@@ -114,9 +114,14 @@ int memcpy_uncached_load_sse41(void *dest, const void *src, size_t n_bytes)
             n -= sizeof(__m128i);
         }
     }
-    _mm_sfence();
+
     if (n)
         memcpy(d, s, n);
+
```
drossetti: @spotluri I guess this comment is incorrect, as when writing to host memory the write fence is needed because of the NT stores.

drossetti:
@spotluri overall, somewhere we should clarify that the available memcpy_to/from_bar cannot be abused to do a BAR-to-BAR copy... the fencing might not be correct

drossetti:
@Akshay-Venkatesh could you please review?

Akshay-Venkatesh:
Will do, Davide.

gdrapi.c
```diff
@@ -350,9 +350,15 @@ int gdr_copy_to_bar(void *bar_ptr, const void *h_ptr, size_t size)
...
             break;
         }
+
         // fall through
         gdr_dbgc(1, "using plain implementation of gdr_copy_to_bar\n");
         memcpy(bar_ptr, h_ptr, size);
+
+        // fencing is needed even for plain memcpy(), due to performance
+        // being hit by delayed flushing of WC buffers
+        _mm_sfence();
```
Akshay-Venkatesh: Not sure if the instruction actually flushes or if it just prevents re-ordering. The description [here](https://msdn.microsoft.com/es-cr/library/5h2w73d1(v=vs.90).aspx) doesn't indicate so. 
For performance to have improved may be the fence has a flush side-effect? Lmk your thoughts

drossetti: Sfence is known to flush the CPU WC buffer. Not easy to find a doc ref though.

PR #37, opened 2018-04-10
gdrdrv: fix use-after-free of variable mr in case of error

daniel-k:
This used to crash the kernel when something went wrong. Maybe the `memset` should be done before freeing the object, but anyway I'm not sure why this should be done.

### DIFF
--- a/gdrdrv/gdrdrv.c
+++ b/gdrdrv/gdrdrv.c
@@ -418,8 +418,8 @@ out:
     }
 
     if (ret && mr) {
-        kfree(mr);
         memset(mr, 0, sizeof(*mr));
+        kfree(mr);
         mr = NULL;
     }
 


drossetti:
@daniel-k I'd rather move the memset() before kfree(). would you be able to refactor your change?

daniel-k:
Sorry for the delay, I somehow missed your first mail. I have now moved the memset before kfree and force-pushed, see 37a2e2f.

Btw: I also added CMake support [over here](https://github.com/daniel-k/gdrcopy/blob/feature/cmake/CMakeLists.txt). If you're interested, I can create a PR =)

drossetti:
merged
@daniel-k thank you

drossetti:
@daniel-k I would really appreciate if you could contact me directly at drossetti@nvidia.com.

PR #46, opened 2019-01-12
Use cached CPU mappings on POWER9

### LINKED ISSUE #29: fix MMIO load/store performance on pcc64le

drossetti:
An alternative performance fix for POWER9. 

For example on a 4 GPU server, with this patch copy bandwidth is roughly 37GB/s

```
$ ./copybw.sh 
+ GDRCOPY_ENABLE_LOGGING=1
+ GDRCOPY_LOG_LEVEL=0
+ LD_LIBRARY_PATH=/home/user/drossetti/Coral/gdrcopy:/home/user/drossetti/Coral/gdrcopy:
+ numactl -N 0 -l ./copybw -d 0 -s 262144 -o 0 -c 262144
GPU id:0 name:Tesla V100-SXM2-16GB PCI domain: 4 bus: 4 device: 0
GPU id:1 name:Tesla V100-SXM2-16GB PCI domain: 4 bus: 5 device: 0
GPU id:2 name:Tesla V100-SXM2-16GB PCI domain: 53 bus: 3 device: 0
GPU id:3 name:Tesla V100-SXM2-16GB PCI domain: 53 bus: 4 device: 0
selecting device 0
testing size: 262144
rounded size: 262144
device ptr: 7fff79600000
bar_ptr: 0x7fffacd80000
info.va: 7fff79600000
info.mapped_size: 262144
info.page_size: 65536
page offset: 0
user-space pointer:0x7fffacd80000
BAR writing test, size=262144 offset=0 num_iters=10000
DBG:  using plain implementation of gdr_copy_to_bar
BAR1 write BW: 37011.3MB/s
BAR reading test, size=262144 offset=0 num_iters=100
DBG:  using plain implementation of gdr_copy_from_bar
BAR1 read BW: 37041MB/s
unmapping buffer
unpinning buffer
closing gdrdrv

$ ./copybw.sh 
+ GDRCOPY_ENABLE_LOGGING=1
+ GDRCOPY_LOG_LEVEL=0
+ LD_LIBRARY_PATH=/home/user/drossetti/Coral/gdrcopy:/home/user/drossetti/Coral/gdrcopy:
+ numactl -N 0 -l ./copybw -d 0 -s 16777216 -o 0 -c 1048576
GPU id:0 name:Tesla V100-SXM2-16GB PCI domain: 4 bus: 4 device: 0
GPU id:1 name:Tesla V100-SXM2-16GB PCI domain: 4 bus: 5 device: 0
GPU id:2 name:Tesla V100-SXM2-16GB PCI domain: 53 bus: 3 device: 0
GPU id:3 name:Tesla V100-SXM2-16GB PCI domain: 53 bus: 4 device: 0
selecting device 0
testing size: 16777216
rounded size: 16777216
device ptr: 7fff79600000
bar_ptr: 0x7fff99960000
info.va: 7fff79600000
info.mapped_size: 16777216
info.page_size: 65536
page offset: 0
user-space pointer:0x7fff99960000
BAR writing test, size=1048576 offset=0 num_iters=10000
DBG:  using plain implementation of gdr_copy_to_bar
BAR1 write BW: 38106.5MB/s
BAR reading test, size=1048576 offset=0 num_iters=100
DBG:  using plain implementation of gdr_copy_from_bar
BAR1 read BW: 37759.4MB/s
unmapping buffer
unpinning buffer
closing gdrdrv
```

thi should fix #29

### DIFF
--- a/common.hpp
+++ b/common.hpp
@@ -63,7 +63,7 @@ static int compare_buf(uint32_t *ref_buf, uint32_t *buf, size_t size)
 {
     int diff = 0;
     if (size % 4 != 0U) {
-        printf("warning: buffer size %d is not dword aligned, ignoring trailing bytes\n", size);
+        printf("warning: buffer size %zu is not dword aligned, ignoring trailing bytes\n", size);
         size -= (size % 4);
     }
     unsigned ndwords = size/sizeof(uint32_t);
--- a/common.hpp
+++ b/common.hpp
@@ -46,9 +46,15 @@
             ASSERT(cudaSuccess == result);     \
         } while (0)
 
+static inline bool operator==(const gdr_mh_t &a, const gdr_mh_t &b) {
+    return a.h == b.h;
+}
+
+static const gdr_mh_t null_mh = {0};
+
 #define ASSERT_EQ(P, V) ASSERT((P) == (V))
 #define CHECK_EQ(P, V) ASSERT((P) == (V))
-#define ASSERT_NEQ(P, V) ASSERT((P) != (V))
+#define ASSERT_NEQ(P, V) ASSERT(!((P) == (V)))
 #define BREAK_IF_NEQ(P, V) if((P) != (V)) break
 #define BEGIN_CHECK do
 #define END_CHECK while(0)
--- a/common.hpp
+++ b/common.hpp
@@ -63,7 +63,7 @@ static int compare_buf(uint32_t *ref_buf, uint32_t *buf, size_t size)
 {
     int diff = 0;
     if (size % 4 != 0U) {
-        printf("warning: buffer size %d is not dword aligned, ignoring trailing bytes\n");
+        printf("warning: buffer size %d is not dword aligned, ignoring trailing bytes\n", size);
         size -= (size % 4);
     }
     unsigned ndwords = size/sizeof(uint32_t);
--- a/gdrapi.h
+++ b/gdrapi.h
@@ -68,6 +68,7 @@ gdr_t gdr_open();
 // gdr_unmap on all mappings before calling gdr_close.
 int gdr_close(gdr_t g);
 
+// The handle to a GPU memory mapping
 typedef struct gdr_mh_s {
   unsigned long h;
 } gdr_mh_t;
--- a/gdrapi.h
+++ b/gdrapi.h
@@ -68,7 +68,7 @@ gdr_t gdr_open();
 // gdr_unmap on all mappings before calling gdr_close.
 int gdr_close(gdr_t g);
 
-// The handle to a GPU memory mapping
+// The handle to a user-space GPU memory mapping
 typedef struct gdr_mh_s {
   unsigned long h;
 } gdr_mh_t;
@@ -99,8 +99,8 @@ struct gdr_info {
     uint32_t page_size;
     uint64_t tm_cycles;
     uint32_t cycles_per_ms;
-    uint32_t mapped;
-    uint32_t wc_mapping;
+    unsigned mapped:1;
+    unsigned wc_mapping:1;
 };
 typedef struct gdr_info gdr_info_t;
 int gdr_get_info(gdr_t g, gdr_mh_t handle, gdr_info_t *info);
@@ -118,10 +118,10 @@ int gdr_map(gdr_t g, gdr_mh_t handle, void **va, size_t size);
 // First invoke gdr_unmap() then gdr_unpin_buffer().
 int gdr_unmap(gdr_t g, gdr_mh_t handle, void *va, size_t size);
 
-// gpubar_ptr is a user-space virtual address, i.e. one returned by gdr_map()
-int gdr_copy_to_bar(void  *gpubar_ptr, const void *cpumem_ptr, size_t size);
-int gdr_copy_from_bar(void *cpumem_ptr, const void *gpubar_ptr, size_t size);
-
+// map_d_ptr is the user-space virtual address belonging to a mapping of a device memory buffer,
+// i.e. one returned by gdr_map()
+int gdr_copy_to_mapping(gdr_mh_t handle, void *map_d_ptr, const void *h_ptr, size_t size);
+int gdr_copy_from_mapping(gdr_mh_t handle, void *h_ptr, const void *map_d_ptr, size_t size);
 
 #ifdef __cplusplus
 }
--- a/gdrapi.h
+++ b/gdrapi.h
@@ -68,9 +68,12 @@ gdr_t gdr_open();
 // gdr_unmap on all mappings before calling gdr_close.
 int gdr_close(gdr_t g);
 
+typedef struct gdr_mh_s {
+  unsigned long h;
+} gdr_mh_t;
+
 // Map device memory buffer on GPU BAR1, returning an handle.
 // Memory is still not accessible to user-space.
-typedef uint32_t gdr_mh_t;
 int gdr_pin_buffer(gdr_t g, unsigned long addr, size_t size, uint64_t p2p_token, uint32_t va_space, gdr_mh_t *handle);
 
 // Unmap the handle. 
--- a/gdrapi.h
+++ b/gdrapi.h
@@ -95,6 +95,8 @@ struct gdr_info {
     uint32_t page_size;
     uint64_t tm_cycles;
     uint32_t cycles_per_ms;
+    uint32_t mapped;
+    uint32_t wc_mapping;
 };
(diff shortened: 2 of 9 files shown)


rsblackmore:
As referenced in the code, the cpu_can_cache_gpu_mappings should depend on whether or not the mapping is through the NPU or PCIe bus.  I think that test has to be done at the time the mapping is done rather than at initialization time. It is certainly possible, but probably not recommended, to have a PCIe attached Volta and a NPU attached Volta on the same system. I believe that at the time of the mapping, the physical address is available and some address bits should determine whether or not it is a cacheable address. I will dig up an example.

drossetti:
> Looks like cpu_can_cache_gpu_mappings is enabled when CONFIG_PPC64 + PVR_POWER9 is detected.
> Is this the only condition under which cpu_can_cache_gpu_mappings can be enabled? If so why not elsewhere?

Enabling cached mappings requires a suitable platform. So far only the POWER9 with GPUs connected via the NPU unit satisfy those requirements.

> What purpose do `nvidia_p2p_*_rsync_*` functions serve? Looks like they're no-ops or error returns. 

Those APIs are exposed by the NVIDIA driver starting from 396+ releases, and they are not related to GPUDirect RDMA.

> What is the purpose of the dummy module?

The side effect of building the dummy module is to create a suitable Module.symvers, which contains the symbol version of the APIs exported by the NVIDIA driver. If everything works correctly, the symbol versions produced here will be the same as those of the real NVIDIA driver.
There are two other, probably better ways of producing a correct Module.symvers, which could be part of a different change.

Note that I have explored using the rsync APIs to detect a P9 machine, but I noticed that then the symbol versions of those functions produced by dummy module do not match those in the real GPU driver. A different strategy would be needed in that case.

rsblackmore:
A simple way to determine if a physical page corresponds to an IO address or a (cacheable) dram address is to use the function page_is_ram. In this case, you could replace
if (!gdrdrv_cpu_can_cache_gpu_mappings)
with 
if (!gdrdrv_cpu_can_cache_gpu_mappings && !page_is_ram(pfn))
page_is_ram is a gpl exported symbol, so the module licence would need to change. At least for the POWER architecture, I believe the function   page_is_ram would suffice for the test and the variable gdrdrv_cpu_can_cache_gpu_mappings would be unnecessary.

rsblackmore:
I should note, that I did test on both P8 and P9 using only the  page_is_ram test and removing the variable  gdrdrv_cpu_can_cache_gpu_mapping variable test. Here I replaced
if (!gdrdrv_cpu_can_cache_gpu_mappings)
with 
if(!page_is_ram(pfn))

Of course, it would be necessary to make sure this is safe on other architectures.
P8:
user-space pointer:0x3effffc70000
BAR writing test, size=131072 offset=0 num_iters=10000
BAR1 write BW: 5257.3MB/s
BAR reading test, size=131072 offset=0 num_iters=100
 BAR1 read BW: 20.3316MB/s

P9:
user-space pointer:0x7fff72c20000
BAR writing test, size=131072 offset=0 num_iters=10000
BAR1 write BW: 38859.1MB/s
BAR reading test, size=131072 offset=0 num_iters=100
BAR1 read BW: 38449.6MB/s

drossetti:
discussed with @Akshay-Venkatesh I'll get rid of the changes to nv-p2p-dummy.c as they are unrelated

rsblackmore:
I don't believe there would be any issue with having the page_is_ram test be under a CONFIG_PPC64 compile time  test if it would be possible to have the PPC64 version be GPL. We know for POWER9 all of the RAM addresses will be 46 bits or less, but that is a HW/FW dependent number and I don't know of a programmatic way of obtaining it.

drossetti:
@rsblackmore I baked the change containing page_is_ram(), which is not GPL-only for PPC64. 
I am still querying for POWER9 regardless, to be on the safe side. 
PLMK if you'd rather remove the pvr_version_is(PVR_POWER9) check altogether.

drossetti:
@e-ago @Akshay-Venkatesh I'll need your +1 on this one, especially Akshay for UCX on P9

Akshay-Venkatesh:
@drossetti Too many things on my plate this week. Can I confirm later this week or is this critical now?

drossetti:
checking again, I can't make my mind... page_is_ram() *IS* GPL only on all platforms, including ppc64le. And in fact, I cannot build this branch on a recent POWER9 machine:
```
[drossetti@pwr04 gdrcopy]$ make
echo "GDRAPI_ARCH=POWER"
GDRAPI_ARCH=POWER
cd gdrdrv; \
make 
make[1]: Entering directory `/pwr05/home/user/drossetti/Coral/gdrcopy/gdrdrv'
Picking NVIDIA driver sources from NVIDIA_SRC_DIR=/usr/src/nvidia-396.64/nvidia. If that does not meet your expectation, you might have a stale driver still around and that might cause problems.
make[2]: Entering directory `/usr/src/kernels/4.14.0-49.18.1.el7a.ppc64le'
  Building modules, stage 2.
  MODPOST 2 modules
FATAL: modpost: GPL-incompatible module gdrdrv.ko uses GPL-only symbol 'page_is_ram'
make[3]: *** [__modpost] Error 1
make[2]: *** [modules] Error 2
make[2]: Leaving directory `/usr/src/kernels/4.14.0-49.18.1.el7a.ppc64le'
make[1]: *** [module] Error 2
make[1]: Leaving directory `/pwr05/home/user/drossetti/Coral/gdrcopy/gdrdrv'
make: *** [driver] Error 2
```

@rsblackmore I know you might not be active on this anymore, but in case, can you suggest someone else who is willing to pick the ball as reviewer here?

drossetti:
assigning to @pakmarkthub, as you are reviewing this already

pakmarkthub:
Does ATS play any role in supporting coherent GPU mapping? If so, is checking with page_is_ram (assume that we can use it) enough to conclude that write-combine is unnecessary? For example, what happen if we disable ATS, which is different from running on non-ATS platforms?

drossetti:
> Does ATS play any role in supporting coherent GPU mapping? If so, is checking with page_is_ram (assume that we can use it) enough to conclude that write-combine is unnecessary? For example, what happen if we disable ATS, which is different from running on non-ATS platforms?

@pakmarkthub GPU use of ATS is not related. GPU can avoid using ATS, still the P9 NPU will bridge traffic between P9 and V100 supporting full CPU memory semantics. 
Besides we have to trust the the powernv kernel arch developers if they decide to return true from page_is_ram().

drossetti:
@pakmarkthub please review

common.hpp
```diff
@@ -57,7 +63,7 @@ static int compare_buf(uint32_t *ref_buf, uint32_t *buf, size_t size)
 {
     int diff = 0;
     if (size % 4 != 0U) {
-        printf("warning: buffer size %d is not dword aligned, ignoring trailing bytes\n");
+        printf("warning: buffer size %d is not dword aligned, ignoring trailing bytes\n", size);
```
pakmarkthub: Suggest to use %zu instead of %d because the type is "size_t"

drossetti: good point, done!

gdrapi.h
```diff
@@ -112,10 +117,10 @@ int gdr_map(gdr_t g, gdr_mh_t handle, void **va, size_t size);
 // First invoke gdr_unmap() then gdr_unpin_buffer().
 int gdr_unmap(gdr_t g, gdr_mh_t handle, void *va, size_t size);
 
-// gpubar_ptr is a user-space virtual address, i.e. one returned by gdr_map()
-int gdr_copy_to_bar(void  *gpubar_ptr, const void *cpumem_ptr, size_t size);
-int gdr_copy_from_bar(void *cpumem_ptr, const void *gpubar_ptr, size_t size);
-
+// map_d_ptr is the user-space virtual address belonging to a mapping of a device memory buffer,
+// i.e. one returned by gdr_map()
+int gdr_copy_to_mapping(gdr_mh_t handle, void *map_d_ptr, const void *h_ptr, size_t size);
```
drossetti: @pakmarkthub One thought I had is that we could in principle maintain some level of source code compatibility by maintaining the old APIs, i.e. `gdr_copy_to_bar()`, while introducing the new APIs, i.e. `gdr_copy_to_mapping`.
The old APIs would not leverage cached mappings, still that could be a good compromise.
Only when using the new APIs the user would experience the full performance improvement due to cached mappings on POWER9.

drossetti:
@pakmarkthub I should have done all the changes you requested, plus I added one design comment. 
Feel free to re-review this PR.

PR #49, opened 2019-04-09
Fixed undefined PVR_POWER9 on linux kernel before 4.12

pakmarkthub:
PVR_POWER9 is defined from linux kernel 4.12 onward. Hence, the code encounters a compilation error when compiling on linux kernel before 4.12. This change should fix the bug.


drossetti:
@pakmarkthub thanks for this fix

PR #51, opened 2019-04-24
Adding test with docker

mario21ic:
Add scripts to run inside a docker container


drossetti:
@pakmarkthub can you please have a look?
@mario21ic 
- I suggest to simply remove 'sudo' from the existing insmod.sh file
- I'd rather add docker specific files in a new specific subdir

Issue #54, opened 2019-05-09
GPU buffer invalidation does not tear-down CPU mappings obtained via gdr_map()


drossetti:
As of today, the nvidia_p2p_get_pages callback does not tear down the CPU mappings created via gdr_map().
This is a potential security threat, as those BAR1 pages could be reused later to expose some other GPU device memory, possibly belonging to a different OS process colocated on the same GPU.

pakmarkthub:
Merged the PR. 

Closing this issue.

Issue #45, opened 2019-01-08
Support for fork


drossetti:
today, forking could lead to spurious prints (https://github.com/NVIDIA/gdrcopy/blob/f54766b21d216584a6839340aa1ebf81a980b235/gdrdrv/gdrdrv.c#L236) and possibly a crash.
tracking here further investigations (a new unit test for this case) and possible mitigations (e.g. CLOEXEC when opening the driver fd).

drossetti:
@pakmarkthub can you help with this?

pakmarkthub:
Merged to the master branch. Closing this issue.

PR #60, opened 2019-06-05
Improve invalidation mechanism and harden security

pakmarkthub:
**Problems:**
- Issue #54: GPU buffer invalidation does not tear-down CPU mappings obtained via gdr_map()
- Issue #45: Support for fork
- Issue #62: fd to /dev/gdrdrv should not be sharable with other processes
- Issue #65: gdr_close does not unmap nor clean up internally alloced gdr_memh_t
- gdr_mh_t.h is generated by random number and is used as mmap offset. Sometimes, the mapping range may overlap another already-mapped range -- imagine the first gdr_mh_t.h is 0x0000 and the map size is 0x0100, then the second gdr_mh_t.h is 0x0020. Although it rarely happens, this bug can lead to undesirable and hard-to-detect behaviors.

**This change:**
- fixes the bug that causes unmap_mapping_range to do nothing. We change the address passed to that API to vm_pgoff instead of vm_start.
- prevents copy of vma when fork. Hence, the child processes cannot access the parent's mappings anymore. This behavior aligns with CUDA since CUDA contexts are not transferred to the child processes.
- checks "struct pid *" to ensure that the caller is the same as the process that opens gdrdrv (calls to gdr_open). Hence, the use of opened gdrdrv fd is local to the creator process.
- separates the mapping address space between each process. Two processes map to different GPU memory even with the same mmap offset.
- makes the current gdr_mh_t.h sit next to the previous mapped range. It also guarantees that no two ranges will be overlapped due to the offsets.

**Presubmit testings:**
- gc04:
   - 4.15.0-50-generic #54-Ubuntu SMP Mon May 6 18:46:08 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
   - P100 GPU
   - Intel(R) Xeon(R) CPU E5-2687W v4 @ 3.00GHz

### DIFF
--- a/gdrdrv/gdrdrv.c
+++ b/gdrdrv/gdrdrv.c
@@ -36,6 +36,7 @@
 #include <linux/timex.h>
 #include <linux/timer.h>
 #include <linux/sched/mm.h>
+#include <linux/sched.h>
 
 #ifndef random_get_entropy
 #define random_get_entropy()	get_cycles()
@@ -240,6 +241,7 @@ struct gdr_info {
     // simple low-performance linked-list implementation
     struct list_head mr_list;
     struct mutex lock;
+    struct pid *pid;
 };
 typedef struct gdr_info gdr_info_t;
 
@@ -272,6 +274,8 @@ static int gdrdrv_open(struct inode *inode, struct file *filp)
 
     INIT_LIST_HEAD(&info->mr_list);
     mutex_init(&info->lock);
+
+    info->pid = task_pid(current);
 
     filp->private_data = info;
 
@@ -642,6 +646,15 @@ static int gdrdrv_ioctl(struct inode *inode, struct file *filp, unsigned int cmd
         return -EINVAL;
     }
 
+    if (!info) {
+        gdr_err("filp contains no info\n");
+        return -EIO;
+    }
+    if (info->pid != task_pid(current)) {
+        gdr_err("filp is not opened by the current process\n");
+        return -EACCES;
+    }
+
     switch (cmd) {
     case GDRDRV_IOC_PIN_BUFFER:
         ret = gdrdrv_pin_buffer(info, argp);
@@ -762,10 +775,13 @@ static int gdrdrv_mmap(struct file *filp, struct vm_area_struct *vma)
 
     gdr_info("mmap filp=0x%p vma=0x%p vm_file=0x%p start=0x%lx size=%zu off=0x%lx\n", filp, vma, vma->vm_file, vma->vm_start, size, vma->vm_pgoff);
 
-    if (info == NULL) {
-        gdr_dbg("gdr_info is not found\n");
-        ret = -EINVAL;
-        goto out;
+    if (!info) {
+        gdr_err("filp contains no info\n");
+        return -EIO;
+    }
+    if (info->pid != task_pid(current)) {
+        gdr_err("filp is not opened by the current process\n");
+        return -EACCES;
     }
 
     handle = gdrdrv_handle_from_off(vma->vm_pgoff);
--- a/gdrdrv/gdrdrv.c
+++ b/gdrdrv/gdrdrv.c
@@ -35,7 +35,6 @@
 #include <linux/io.h>
 #include <linux/timex.h>
 #include <linux/timer.h>
-#include <linux/sched/mm.h>
 #include <linux/sched.h>
 
 #if LINUX_VERSION_CODE <= KERNEL_VERSION(2,6,32)
--- a/gdrdrv/gdrdrv.c
+++ b/gdrdrv/gdrdrv.c
@@ -258,7 +258,7 @@ static int gdrdrv_open(struct inode *inode, struct file *filp)
     int ret = 0;
     gdr_info_t *info = NULL;
 
-    gdr_dbg("minor=%d filep=0x%p\n", minor, filp);
+    gdr_dbg("minor=%d filep=0x%px\n", minor, filp);
     if(minor >= 1) {
         gdr_err("device minor number too big!\n");
         ret = -ENXIO;
@@ -307,7 +307,7 @@ static int gdrdrv_release(struct inode *inode, struct file *filp)
     mutex_lock(&info->lock);
     list_for_each_safe(p, n, &info->mr_list) {
         mr = list_entry(p, gdr_mr_t, node);
-        gdr_info("freeing MR=0x%p\n", mr);
+        gdr_info("freeing MR=0x%px\n", mr);
         if (gdr_mr_is_mapped(mr)) {
             mutex_unlock(&info->lock);
             gdr_mr_destroy_all_mappings(mr);
@@ -405,7 +405,7 @@ static int gdrdrv_pin_buffer(gdr_info_t *info, void __user *_params)
     cycles_t ta, tb;
 
     if (copy_from_user(&params, _params, sizeof(params))) {
-        gdr_err("copy_from_user failed on user pointer %p\n", _params);
+        gdr_err("copy_from_user failed on user pointer 0x%px\n", _params);
         ret = -EFAULT;
         goto out;
     }
@@ -526,7 +526,7 @@ out:
     }
 
     if (!ret && copy_to_user(_params, &params, sizeof(params))) {
-        gdr_err("copy_to_user failed on user pointer %p\n", _params);
+        gdr_err("copy_to_user failed on user pointer 0x%px\n", _params);
         ret = -EFAULT;
     }
 
@@ -542,7 +542,7 @@ static int gdrdrv_unpin_buffer(gdr_info_t *info, void __user *_params)
     gdr_mr_t *mr = NULL;
 
     if (copy_from_user(&params, _params, sizeof(params))) {
-        gdr_err("copy_from_user failed on user pointer %p\n", _params);
+        gdr_err("copy_from_user failed on user pointer 0x%px\n", _params);
         return -EFAULT;
     }
 
@@ -555,7 +555,7 @@ static int gdrdrv_unpin_buffer(gdr_info_t *info, void __user *_params)
         ret = -EINVAL;
     } else {
         if (gdr_mr_is_mapped(mr)) {
-            gdr_err("trying to unpin mapped mr 0x%p\n", mr);
+            gdr_err("trying to unpin mapped mr 0x%px\n", mr);
             ret = -EBUSY;
         } else {
             list_del(&mr->node);
@@ -591,7 +591,7 @@ static int gdrdrv_get_cb_flag(gdr_info_t *info, void __user *_params)
     gdr_mr_t *mr = NULL;
 
     if (copy_from_user(&params, _params, sizeof(params))) {
-        gdr_err("copy_from_user failed on user pointer %p\n", _params);
+        gdr_err("copy_from_user failed on user pointer 0x%px\n", _params);
         return -EFAULT;
     }
     mr = gdr_mr_from_handle(info, params.handle);
@@ -604,7 +604,7 @@ static int gdrdrv_get_cb_flag(gdr_info_t *info, void __user *_params)
     params.flag = !!mr->cb_flag;
 
     if (copy_to_user(_params, &params, sizeof(params))) {
-        gdr_err("copy_to_user failed on user pointer %p\n", _params);
+        gdr_err("copy_to_user failed on user pointer 0x%px\n", _params);
         ret = -EFAULT;
     }
  out:
@@ -620,7 +620,7 @@ static int gdrdrv_get_info(gdr_info_t *info, void __user *_params)
     gdr_mr_t *mr = NULL;
 
     if (copy_from_user(&params, _params, sizeof(params))) {
-        gdr_err("copy_from_user failed on user pointer %p\n", _params);
+        gdr_err("copy_from_user failed on user pointer 0x%px\n", _params);
         ret = -EFAULT;
         goto out;
     }
@@ -640,7 +640,7 @@ static int gdrdrv_get_info(gdr_info_t *info, void __user *_params)
     params.mapped      = gdr_mr_is_mapped(mr);
     params.wc_mapping  = gdr_mr_is_wc_mapping(mr);
     if (copy_to_user(_params, &params, sizeof(params))) {
-        gdr_err("copy_to_user failed on user pointer %p\n", _params);
+        gdr_err("copy_to_user failed on user pointer 0x%px\n", _params);
         ret = -EFAULT;
     }
  out:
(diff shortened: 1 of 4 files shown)


pakmarkthub:
@Akshay-Venkatesh Can I ask you to review this PR as well? I would like to make sure that it does not break UCX.

drossetti:
@pakmarkthub regarding
>>> gdr_mh_t.h is generated by random number and is used as mmap offset. Sometimes, the mapping range may overlap another already-mapped range -- imagine the first gdr_mh_t.h is 0x0000 and the map size is 0x0100, then the second gdr_mh_t.h is 0x0020. Although it rarely happens, this bug can lead to undesirable and hard-to-detect behaviors.

I don't fully follow. Handles are uniquely associated to a single call to nvidia_p2p_get_pages. In that respect they are references to instances of kernel-side struct nvidia p2p page table. They don't have the concept of spanning a virtual address range.

pakmarkthub:
@drossetti As discussed offline, unmap_mapping_range unmaps all mappings based on the specified file offset and range (see https://elixir.bootlin.com/linux/v4.15.18/source/mm/memory.c#L2811). If we have two mappings that have overlapped file-offset range, unmap_mapping_range will remove both of them. 

An example undesirable situation:
- Do cuMemAlloc(&d_A, ...) and cuMemAlloc(&d_B, ...).
- Do gdr_pin_buffer for both d_A and d_B. The handle numbers returned from gdrdrv are very close. This situation rarely occurs but random_get_entropy() cannot guarantee that it will **never** occur.
- Do gdr_map(..., &bar_ptr_A, ...) on d_A, and gdr_map(..., &bar_ptr_B, ...) on d_B.
- Call cuMemFree(d_A).
- This API triggers call back from nvidia_p2p_get_pages to destroy the bar_ptr_A mapping.
- gdrdrv calls unmap_mapping_range with the offset based on handle_A.
- unmap_mapping_range removes bar_ptr_B mapping as well because the offset based on handle_B is between handle_A and handle_A + size_A, which is a bug.

gdrdrv/gdrdrv.c
```diff
@@ -479,16 +493,23 @@ static int gdrdrv_pin_buffer(gdr_info_t *info, void __user *_params)
...
-
     // here a typical driver would use the page_table to fill in some HW
     // DMA data structure
 
-    params.handle = mr->handle;
-
     mutex_lock(&info->lock);
     list_add(&mr->node, &info->mr_list);
+
+    // The user treats handle as the offset value for mmap. To guarantee no
```
drossetti: What about something like:
// the user-space library passes the memory handle as the mmap offset,
// and offsets are used to determine the VMAs to delete during invalidation.
// Hence, we need [handle,handle+size-1] to correspond to a unique VMA.
// Note that size here must match the original mmap size

gdrdrv/gdrdrv.c
```diff
@@ -811,7 +856,7 @@ static int gdrdrv_mmap(struct file *filp, struct vm_area_struct *vma)
     // this also works as the mapped flag for this mr
     mr->cpu_mapping_type = GDR_MR_CACHING;
     vma->vm_ops = &gdrdrv_vm_ops;
-    gdr_dbg("overwriting vma->vm_private_data=%px with mr=%px\n", vma->vm_private_data, mr);
+    gdr_dbg("overwriting vma->vm_private_data=0x%p with mr=0x%p\n", vma->vm_private_data, mr);
```
drossetti: note that %px is not the same as 0x%p, check the kernel doc

test_invalidation.cpp
```diff
@@ -0,0 +1,1138 @@
+/*
+ * Copyright (c) 2014, NVIDIA CORPORATION. All rights reserved.
```
drossetti: 2019

test_invalidation.cpp
```diff
@@ -0,0 +1,1138 @@
...
+    print_dbg("%s: Start\n", myname);
+
+    if (pid == 0) {
+        close(filedes_0[0]);
+        close(filedes_1[1]);
+
+        read_fd = filedes_1[0];
+        write_fd = filedes_0[1];
+
+        srand(rand());
```
drossetti: Why is srand() need?
Also, I'm not sure srand(rand()) makes sense.

test_invalidation.cpp
```diff
@@ -0,0 +1,1138 @@
...
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <iostream>
+#include <cuda.h>
+#include <cuda_runtime_api.h>
+#include <check.h>
```
drossetti: you need to capture the dependency on check somehow, e.g. in configure and dep/rpm

gdrdrv/gdrdrv.c
```diff
@@ -479,16 +493,23 @@ static int gdrdrv_pin_buffer(gdr_info_t *info, void __user *_params)
...
-    params.handle = mr->handle;
-
     mutex_lock(&info->lock);
     list_add(&mr->node, &info->mr_list);
+
+    // The user treats handle as the offset value for mmap. To guarantee no
+    // overlapping, the next offset will sit next to this mapping range.
+    // Remember that the address space is local to the process. Thus, two
+    // processes do mmap with the same offset get different mappings.
+    mr->handle = info->next_offset;
```
drossetti: can we move this code to an helper function?
besides, there is a wrap-around problem in this design, isn't there?
it should be at least documented
lacking a better solution, you should return an error when wrapping

gdrdrv/gdrdrv.c
```diff
@@ -293,7 +307,7 @@ static int gdrdrv_release(struct inode *inode, struct file *filp)
     mutex_lock(&info->lock);
     list_for_each_safe(p, n, &info->mr_list) {
```
drossetti: do you need the info->pid != task_pid(current) check here?

pakmarkthub:
@drossetti Please review

gdrdrv/gdrdrv.c
```diff
@@ -378,6 +433,39 @@ static void gdrdrv_get_pages_free_callback(void *data)
...
+    // Hence, we need [(handle << PAGE_SHIFT), (handle << PAGE_SHIFT) + size - 1] 
+    // to correspond to a unique VMA.  Note that size here must match the
+    // original mmap size
+
+    gdr_hnd_t next_handle = info->next_handle + (mr->mapped_size >> PAGE_SHIFT);
+
+    // The end of this range is wrapped around. The calculation uses PAGE_SHIFT
+    // because handle is equal vm_pgoff (aka file offset << PAGE_SHIFT). We
+    // indicate this case as -ENOMEM error.
+    if ((next_handle & ((gdr_hnd_t)(-1) >> PAGE_SHIFT)) < info->next_handle)
```
drossetti: I don't like failing eagerly. It should fail when next handle is going to be used.

gdrdrv/gdrdrv.c
```diff
@@ -479,16 +566,20 @@ static int gdrdrv_pin_buffer(gdr_info_t *info, void __user *_params)
...
     // DMA data structure
 
-    params.handle = mr->handle;
-
     mutex_lock(&info->lock);
     list_add(&mr->node, &info->mr_list);
+
+    if (gdr_generate_mr_handle(info, mr) != 0) {
+        gdr_err("No address space left for future BAR1 mapping.\n");
+        ret = -ENOMEM;
```
drossetti: same here, no point in failing when there is still address space for this mapping.

gdrdrv/gdrdrv.c
```diff
@@ -479,16 +566,20 @@ static int gdrdrv_pin_buffer(gdr_info_t *info, void __user *_params)
...
-
     // here a typical driver would use the page_table to fill in some HW
     // DMA data structure
 
-    params.handle = mr->handle;
-
     mutex_lock(&info->lock);
     list_add(&mr->node, &info->mr_list);
+
+    if (gdr_generate_mr_handle(info, mr) != 0) {
```
drossetti: why is this call placed here? it returns an error but keeps the mr in the list.
it should add mr to the list only when it is ok.

pakmarkthub:
@drossetti Thank you. I have fixed the bug.

Makefile
```diff
@@ -86,6 +87,9 @@ validate: validate.o $(LIB)
 copybw: copybw.o $(LIB)
 	$(LINK.cc)  -o $@ $^ $(LIBS)
 
+test_invalidation: test_invalidation.o $(LIB)
+	$(LINK.cc)  -o $@ $^ -lcheck -lcudart -lcuda -lpthread -lrt -lsubunit
```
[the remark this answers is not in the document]
pakmarkthub: At least on Ubuntu 18.04 with libcheck installed from "apt install check", I cannot compile without "-lsubunit". This seems to be a dependency from libcheck.a itself. The error is as below:

```
g++ -O2 -I /usr/local/cuda/include -I gdrdrv/ -I /usr/local/cuda/include -D GDRAPI_ARCH=X86 -L /usr/local/cuda/lib64 -L /usr/local/cuda/lib -L /usr/lib64/nvidia -L /usr/lib/nvidia -L /usr/local/cuda/lib64   -o sanity sanity.o libgdrapi.so.1.2 -lcheck -lcudart -lcuda -lpthread -lrt
/usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libcheck.a(check_log.o): In function `subunit_lfun':
(.text+0x5f4): undefined reference to `subunit_test_start'
(.text+0x6bf): undefined reference to `subunit_test_fail'
(.text+0x6d4): undefined reference to `subunit_test_pass'
(.text+0x6ef): undefined reference to `subunit_test_error'
collect2: error: ld returned 1 exit status
```

drossetti:
@pakmarkthub 
is the comment in gdrapi.h copied below still true ?
```
// Note that altough BAR mappings of GPU memory are destroyed, user-space                                                                                                                                                                                                       
// mappings are not. So therefore user code is responsible of calling                                                                                                                                                                                                           
// gdr_unmap on all mappings before calling gdr_close.                                                                                                                                                                                                                          
int gdr_close(gdr_t g);
```

pakmarkthub:
@drossetti Thanks Davide. This revision is ready for review.

pakmarkthub:
@e-ago Could you also take a look? I want to make sure that this PR does not break what you are working on.

pakmarkthub:
Please help reviewing this PR. It is one of the blockers for the automated test.

gdrdrv/gdrdrv.c
```diff
@@ -872,14 +993,16 @@ static int gdrdrv_mmap(struct file *filp, struct vm_area_struct *vma)
...
     if (ret) {
-        mr->vma = NULL;
-        mr->mapping = NULL;
-        mr->cpu_mapping_type = GDR_MR_NONE;
-        // TODO: tear down stale partial mappings
+        if (mr) {
+            mr->vma = NULL;
+            mr->mapping = NULL;
+            mr->cpu_mapping_type = GDR_MR_NONE;
+            // TODO: tear down stale partial mappings
```
drossetti: This comment might have never been true, i.e. in case of error, the vma is destroyed and all partial mappings are automatically removed. 
Still we should verify that.

sanity.cpp
```diff
@@ -0,0 +1,1385 @@
...
+        print_dbg("recvmsg failed no data");
+        return -1;
+    }
+
+    cmsg = CMSG_FIRSTHDR(&msg);
+    memmove(&fd, CMSG_DATA(cmsg), sizeof(int));
+    return fd;
+}
+
+typedef struct gdr_memh_t {
```
drossetti: rather than duplicating these type definitions here, I'd rather have a shared header file.

sanity.cpp
```diff
@@ -0,0 +1,1385 @@
...
+
+    init_hbuf_walking_bit(init_buf, size);
+    memset(copy_buf, 0, sizeof(*copy_buf) * sizeof(uint32_t));
+
+    gdr_t g = gdr_open();
+    ASSERT_NEQ(g, (void*)0);
+
+    gdr_mh_t mh = {0};
+    if (mh == null_mh) 
+		print_dbg("miao\n");
```
drossetti: this looks like stale code

sanity.cpp
```diff
@@ -0,0 +1,1385 @@
...
+	CUdeviceptr d_ptr = d_A;
```
drossetti: indentation problem

sanity.cpp
```diff
@@ -0,0 +1,1385 @@
...
+    uint32_t *copy_buf = new uint32_t[size];
+
+    init_hbuf_walking_bit(init_buf, size);
+    memset(copy_buf, 0xA5, size * sizeof(*copy_buf));
+
+    gdr_t g = gdr_open();
+    ASSERT_NEQ(g, (void*)0);
+
+    gdr_mh_t mh;
+
```
drossetti: noticed the weird indentation... are you missing curly braces here?

gdrdrv/gdrdrv.c
```diff
@@ -314,6 +367,8 @@ static int gdrdrv_release(struct inode *inode, struct file *filp)
     }
     mutex_unlock(&info->lock);
 
+    filp->f_mapping = NULL;
```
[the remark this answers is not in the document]
pakmarkthub: address_space_init_once seems to be an asymmetric API. No deinit is available. I guess that because address_space_init_once does not do memory allocation, the deinit API is not required.

Also,  setting filp->f_mapping back to the original mapping should be unnecessary. This is because filp is being destroyed after gdrdrv_release returns.

This method is inspired by nvidia-uvm. Please refer to the uvm_release function in uvm8.c.

gdrdrv/gdrdrv.c
pakmarkthub: Confirmed that if gdrdrv_mmap returns an error, Linux kernel will clean up any partial mappings for us. See https://elixir.bootlin.com/linux/latest/source/mm/mmap.c#L1788 for reference.

pakmarkthub:
@drossetti Thank you for the comments. I have fixed the issues. Please re-review the PR.

gdrapi_internal.h
```diff
@@ -0,0 +1,12 @@
+typedef struct gdr_memh_t {
```
drossetti: 3 missing things:
- License header
- protection from multiple inclusion
- extern "C"

drossetti:
The  sanity test is a fairly large piece of code. Would it be possible to move that to a separate PR?
Or at least fix the PR cover letter which still refers to "test_validation.cpp"

pakmarkthub:
@drossetti As requested, I separated out the sanity test suite to PR #71. This PR now focuses on the invalidation issues only. Please re-review.

gdrdrv/gdrdrv.c
```diff
@@ -479,16 +573,21 @@ static int gdrdrv_pin_buffer(gdr_info_t *info, void __user *_params)
...
+    if (gdr_generate_mr_handle(info, mr) != 0) {
+        gdr_err("No address space left for BAR1 mapping.\n");
+        ret = -ENOMEM;
+    }
+
+    if (!ret)
+        list_add(&mr->node, &info->mr_list);
     mutex_unlock(&info->lock);
 
+    params.handle = mr->handle;
```
drossetti: purely for style, I'd rather access mr->handle when it has a sane value only, i.e. when there are no errors so after list_add().

drossetti:
merged

PR #57, opened 2019-05-29
Switch to use autotools

pakmarkthub:
**This change:**
- addresses issue #55.
- addresses issue #47.
- removes handcrafted Makefile and switches to use autotools.
- introduces autogen.sh for setting up autotools in the repository.
- changes build_packages.sh and gdrcopy.spec to build with autotools.
- removes unused / obsoleted config_arch.

**Presubmit Testings:**
- configure and build_packages.sh on Ubuntu18.04 (4.15.0-45-generic; x86_64)
- configure and build_packages.sh on Red Hat Enterprise Linux Server 7.5 (ppc64le)
- configure and build_packages.sh on Red Hat Enterprise Linux Workstation 6.6 (2.6.32-642.el6.x86_64; x86_64)

PR #70, opened 2019-07-14
Update Makefile

metere1llnl:
Allows for local installation. This is useful when compiling gdrcopy on a cluster or wherever the user does not have rights to do global installations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,12 @@ exes: $(EXES)
 
 install: lib_install #drv_install
 
-lib_install:
+makedirs:
+        @ echo "creating $(DESTLIB) ..."
+        mkdir -p $(DESTLIB)
+        mkdir -p $(PREFIX)/include
+
+lib_install: makedirs
 	@ echo "installing in $(PREFIX)..." && \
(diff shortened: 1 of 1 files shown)


drossetti:
@metere1llnl we are moving the build & install system to autoconf/automake.
We would appreciate your testing of PR #57

metere1llnl:
Would BLT (https://github.com/LLNL/blt) be a better alternative in your opinion?

drossetti:
@metere1llnl I'm not familiar with that. 
GDRcopy is a small library with few dependencies, so I would rather stick to the GNU build system.

Makefile
```diff
@@ -45,7 +45,12 @@ exes: $(EXES)
 
 install: lib_install #drv_install
 
-lib_install:
+makedirs:
+        @ echo "creating $(DESTLIB) ..."
+        mkdir -p $(DESTLIB)
```
drossetti: isn't /usr/bin/install supposed to create directories when necessary? on my RHEL6:
```
$ install --help
...
-D                  create all leading components of DEST except the last,
                        then copy SOURCE to DEST
```

drossetti (changes requested):
is this change still needed?

Issue #68, opened 2019-06-20
provide a run-time version query mechanism


drossetti:
We might consider a run-time query mechanism, like `gdr_query_version(int *major, int *minor)` or the more generic `gdr_get_attribute(int attr, int *value)`,  which would complement the dynamic link time mechanism offered by ld.so.

That would be especially useful, say in MPI libraries, when dynamically loading the library with `dlopen("libgdrapi.so")` and resolving symbols with `dlsym()`, to enforce a run-time compatibility check.

pakmarkthub:
The PR has been merged. Close this issue

Issue #67, opened 2019-06-20
enforce ABI compatibility between user and kernel space components


drossetti:
Currently there is no run-time ABI compatibility check between libgdrapi and gdrdrv. 

That can generate obscure errors, say in a container when libgdapi version A tries to work with baremetal gdrdrv version B.

A possible plan would be:
- to introduce the concept of ABI version in gdrdrv
- to add a new IOCTL to return that version to user-space
- in gdr_open(), check ABI compatibility

pakmarkthub:
Merged the PR. 

Closing this issue.

PR #73, opened 2019-07-31
Version checking and querying

pakmarkthub:
This change:
- addresses issue #67.
- addresses issue #68.
- introduces gdr_runtime_get_version API for querying the libgdrapi version.
- introduces gdr_driver_get_version API for querying the gdrdrv driver version.
- enforces the minimum required version checking between gdrapi annd gdrdrv in gdr_open.


bureddy:
@pakmarkthub @drossetti @spotluri @Akshay-Venkatesh  
What is timeline for 1.4 release. 
currently UCX is broken with master branch because of API change.  
Just wondering if UCX has to query API version to support both new and old APIs (or) just drop the old API and use new API

pakmarkthub:
@bureddy Thank you for reporting this issue. We are targeting the end of this week for internal testing. The actual release date would be shortly after.

In the meantime, can you use version 1.3 when compiling UCX?

Issue #66, opened 2019-06-20
consolidate API versioning


drossetti:
ATM there are 3 places where the library major and minor version are specified:
- gdrapi.h
- Makefile
- gdrcopy.spec
There should be a single place where those version numbers are maintained.

drossetti:
@pakmarkthub that would depend on your autotool PR

pakmarkthub:
Merged the PR. 

Closing this issue.

Issue #56, opened 2019-05-29
use dkms in gdrdrv


drossetti:
dkms is needed anyway for deb kernel module packages

pakmarkthub:
Merged the PR. 

Closing this issue.

Issue #31, opened 2018-03-09
support creation of deb package


drossetti:

drossetti:
@pakmarkthub please see @e-ago seminal work in #32

pakmarkthub:
Merged the PR. 

Closing this issue.

PR #74, opened 2019-08-05
Packaging based on Makefile

pakmarkthub:
**This change:**
- addresses issue #31, #56, and #66.
- moved packaging-related files to under the packages folder.
- created "build-deb-packages.sh" script to automatically create deb packages for libgdrapi and gdrdrv dkms.
- renames "build-packages.sh" to "build-rpm-packages.sh".
- modifies both build-*-packages.sh scripts to be able to run from anywhere.

Presubmit Testing:
- gc03: Ubuntu18.04 (4.15.0-45-generic; x86_64)


packages/debian/README.Debian
```diff
@@ -0,0 +1,228 @@
+gdrcopy for Debian
```
drossetti: would it be possible to generate README.Debian from README.md ?

packages/dkms.conf
```diff
@@ -0,0 +1,5 @@
+PACKAGE_NAME="gdrdrv"
+PACKAGE_VERSION="@VERSION@"
+BUILT_MODULE_NAME[0]="gdrdrv"
+DEST_MODULE_LOCATION[0]="/kernel/drivers/video/fbdev/nvidia/"
```
drossetti: That does not look like a good location. 
What is the standard for add-on modules ?

pakmarkthub: From dkms manpage:

```
DEST_MODULE_LOCATION[#]=
              This directive specifies the destination where a module  should  be  installed  to,
              once  compiled.   It also is used for finding original_modules.  This is a required
              directive, except as noted below. This directive must start with the text "/kernel"
              which  is  in reference to /lib/modules/<kernelversion>/kernel.  Note that for each
              module within a dkms package, the numeric value of # must be the same for  each  of
              BUILT_MODULE_NAME, BUILT_MODULE_LOCATION, DEST_MODULE_NAME and DEST_MODULE_LOCATION
              and     that     the      numbering      should      start      at      0      (eg.
              DEST_MODULE_LOCATION[0]="/kernel/drivers/something/"
              DEST_MODULE_LOCATION[1]="/kernel/drivers/other/").

              DEST_MODULE_LOCATION is ignored on Fedora Core 6 and  higher,  Red  Hat  Enterprise
              Linux  5 and higher, Novell SuSE Linux Enterprise Server 10 and higher, Novell SuSE
              Linux 10.0 and  higher,  and  Ubuntu.  Instead,  the  proper  distribution-specific
              directory is used.
```

NVIDIA driver installs the modules in `/lib/modules/<kver>/kernel/drivers/video/`.
mlx5*.ko are found in two locations:
- `/lib/modules/<kver>/kernel/drivers/infiniband/hw/mlx5/`.
- `/lib/modules/<kver>/kernel/drivers/net/ethernet/mellanox/mlx5/core/`

An example in https://wiki.archlinux.org/index.php/Dynamic_Kernel_Module_Support suggests `/lib/modules/<kver>/kernel/drivers/misc`.

An example in https://wiki.debian.org/KernelDKMS uses `/lib/modules/<kver>/extra/`.

Those examples do not even follow the dkms manpage. 

I do not know the best practice, but technically I think that anywhere inside `/lib/modules/<kver>/` is fine. However, our rpm and deb packages should install gdrdrv.ko in the same location -- just realize that they are different now.

Any suggestion? Should we put gdrdrv.ko in the same location as nvidia*.ko or in a different location?

pakmarkthub:
Addressed the review comments. 

Changed the gdrdrv.ko install location to /lib/modules/<kver>/kernel/drivers/misc/.

@drossetti Please have a look.

Issue #80, opened 2019-08-22
Installing/Uninstalling gdrcopy-kmod.rpm failed on RHEL6


pakmarkthub:
systemctl is used in installation and uninstallation scripts for gdrcopy-kmod.rpm, but RHEL6 does not have systemctl.

PR #81, opened 2019-08-22
Changed "systemctl" to "service" in gdrcopy.spec

pakmarkthub:
Problems:
- Issue #80.

This change:
- modifies the pre and post scripts in gdrcopy.spec to use "service" instead of "systemctl".
- checks for the existence of systemctl and does systemctl daemon-reload if found in the post-uninstall script. This step is needed to suppress systemctl error/warning message when changing any services.

### DIFF
--- a/packages/gdrcopy.spec
+++ b/packages/gdrcopy.spec
@@ -76,17 +76,19 @@ fi
 
 /sbin/chkconfig --add gdrcopy
 
-/usr/bin/systemctl start gdrcopy
+service gdrcopy start
 
 %preun %{kmod}
-/usr/bin/systemctl stop gdrcopy
+service gdrcopy stop
 %{MODPROBE} -rq gdrdrv
 if ! ( /sbin/chkconfig --del gdrcopy > /dev/null 2>&1 ); then
    true
 fi              
 
 %postun %{kmod}
-/usr/bin/systemctl daemon-reload
+if [ -e /usr/bin/systemctl ]; then
+    /usr/bin/systemctl daemon-reload
+fi
 
 
 %clean


pakmarkthub:
@drossetti I came up with this recipe based on how we manually install and uninstall gdrdrv. I believe that most parts are self-explainatory -- copy gdrdrv to the specified location and start gdrcopy service during installation, etc. 

For `systemctl daemon-reload`, systemctl shows this suggestion whenever I remove the service and then call systemctl for any reasons. According to the man pages (see below), this is a way to clean stale configurations after removing services.

From systemctl man page:
```
       daemon-reload
           Reload the systemd manager configuration. This will rerun all generators (see systemd.generator(7)), reload all unit files, and recreate the entire dependency tree. While the daemon is being
           reloaded, all sockets systemd listens on behalf of user configuration will stay accessible.
```

From systemd.generator man page:
```
       Generators are small executables that live in /lib/systemd/system-generators/ and other directories listed above.  systemd(1) will execute those binaries very early at bootup and at configuration
       reload time — before unit files are loaded. Generators may dynamically generate unit files (regular ones, instances as well as templates) and unit file .d/ drop-ins, or create symbolic links to unit
       files to add additional dependencies or instantiate existing templates, thus extending or overriding existing definitions. Their main purpose is to convert configuration files that are not native
       unit files dynamically into native unit files.

       Generators are loaded from a set of paths determined during compilation, as listed above. System and user generators are loaded from directories with names ending in system-generators/ and
       user-generators/, respectively. Generators found in directories listed earlier override the ones with the same name in directories lower in the list. A symlink to /dev/null or an empty file can be
       used to mask a generator, thereby preventing it from running. Please note that the order of the two directories with the highest priority is reversed with respect to the unit load path, and
       generators in /run overwrite those in /etc.

       After installing new generators or updating the configuration, systemctl daemon-reload may be executed. This will delete the previous configuration created by generators, re-run all generators, and
       cause systemd to reload units from disk. See systemctl(1) for more information.
```

drossetti:
I mean in particular the daemon-reload aspect of your change.
How do other packages deal with this very same problem?
Can you for example check nvidia_peer_mem ?

pakmarkthub:
nvidia_peer_mem does not take care of this problem. I believe that it would just annoy the users but does not have negative side-effect.

```
$ sudo rpm -e nvidia_peer_memory
$ systemctl status nv_peer_mem
● nv_peer_mem.service - LSB: Activates/Deactivates nv_peer_mem module to start at boot time.
   Loaded: loaded (/etc/rc.d/init.d/nv_peer_mem; bad; vendor preset: disabled)
   Active: failed (Result: exit-code) since Tue 2019-05-28 17:10:27 PDT; 3 months 9 days ago
     Docs: man:systemd-sysv-generator(8)

May 28 17:10:26 pwr04.nvidia.com systemd[1]: Starting LSB: Activates/Deactivates nv_peer_mem module to start at boot time....
May 28 17:10:27 pwr04.nvidia.com nv_peer_mem[17248]: starting... modprobe: ERROR: could not insert 'nv_peer_mem': Invalid argument
May 28 17:10:27 pwr04.nvidia.com nv_peer_mem[17248]: Failed to load nv_peer_mem
May 28 17:10:27 pwr04.nvidia.com systemd[1]: nv_peer_mem.service: control process exited, code=exited status=1
May 28 17:10:27 pwr04.nvidia.com systemd[1]: Failed to start LSB: Activates/Deactivates nv_peer_mem module to start at boot time..
May 28 17:10:27 pwr04.nvidia.com systemd[1]: Unit nv_peer_mem.service entered failed state.
May 28 17:10:27 pwr04.nvidia.com systemd[1]: nv_peer_mem.service failed.
Warning: nv_peer_mem.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```

drossetti:
make sense, thanks

pakmarkthub:
@drossetti 

In case you don't like daemon-reload, rpm also has scriptlets for properly dealing with systemd (https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd). 

The change is here: https://github.com/pakmarkthub/gdrcopy/commit/8822a1d075096bf8d4e60145ff055d4c74cb8fbb

I have tested the change on both RHEL6 and 7. It works on both systems.

Issue #82, opened 2019-08-27
Version mismatch between modinfo gdrdrv and dpkg -l gdrdrv-dkms


pakmarkthub:
Code from the master branch (bf4848f9c9bc8f06ab40b670620c363864ecbe6c).

```
$ dpkg -l gdrdrv-dkms
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                                           Version                      Architecture                 Description
+++-==============================================-============================-============================-==================================================================================================
ii  gdrdrv-dkms:amd64                              2.0                          amd64                        gdrdrv driver in DKMS format.
```

```
$ modinfo gdrdrv
filename:       /lib/modules/4.15.0-58-generic/updates/dkms/gdrdrv.ko
version:        1.1
description:    GDRCopy kernel-mode driver
license:        MIT
author:         drossetti@nvidia.com
srcversion:     D5FB5F3108420043522DCAC
depends:        nv-p2p-dummy
retpoline:      Y
name:           gdrdrv
vermagic:       4.15.0-58-generic SMP mod_unload 
parm:           dbg_enabled:enable debug tracing (int)
parm:           info_enabled:enable info tracing (int)
```

PR #83, opened 2019-08-27
Changed MODULE_VERSION in gdrdrv to follow GDRDRV_VERSION

pakmarkthub:
Problem:
- Issue #82.

This PR:
- modifies MODULE_VERSION in gdrdrv from hard-code to GDRDRV_*_VERSION.
- the idea about stringify preprocessor values comes from https://stackoverflow.com/questions/240353/convert-a-preprocessor-token-to-a-string.


drossetti:
Love it, thanks

drossetti:
actually STRINGIFY might be common, so I suggest to protect it with  #if !defined

pakmarkthub:
Decided to add _GDRDRV_ prefix to those two preprocessor functions. This is because we cannot guarantee that other _STRINGIFY would have the same syntax as we expected.

pakmarkthub:
@drossetti Please re-review the PR.

src/gdrdrv/gdrdrv.h
```diff
@@ -23,9 +23,13 @@
 #ifndef __GDR_DRV_H__
 #define __GDR_DRV_H__
 
+#define _GDRDRV_STRINGIFY(s)           #s
+#define _GDRDRV_TOSTRING(s)            _GDRDRV_STRINGIFY(s)
```
drossetti: usually identifiers starting with underscore are considered "reserved".
can you get rif of it?

pakmarkthub:
@drossetti I have removed the underscore.

Issue #52, opened 2019-05-09
cudaMalloc can no longer guarantee to return 64kB aligned address


e-ago:
GDRDRV needs 64kB aligned addresses. 
```
gdrdrv_pin_buffer() {
...
    page_virt_start  = params.addr & GPU_PAGE_MASK;
    page_virt_end    = params.addr + params.size - 1;
    rounded_size     = page_virt_end - page_virt_start + 1;
    mr->offset       = params.addr & GPU_PAGE_OFFSET;
...
}
```

and

```
gdrdrv_mmap() {
...
    if (mr->offset) {
        gdr_dbg("offset != 0 is not supported\n");
        ret = -EINVAL;
        goto out;
    }
...
}
```
This is no more guaranteed with the cudaMalloc in recent CUDA drivers (since 410). A temporary WAR could be (at application level) to allocate with the cudaMalloc a memory area that is `size  + GPU_PAGE_SIZE` and then search for the first 64kB aligned address. Something like:
```
alloc_size = (buffer_size + GPU_PAGE_SIZE) & GPU_PAGE_MASK;
cuMemAlloc(&dev_addr, alloc_size);
if(dev_addr % GPU_PAGE_SIZE) {
    dev_addr += (GPU_PAGE_SIZE - (dev_addr % GPU_PAGE_SIZE));
}
```

pakmarkthub:
I also encounter this bug.

@drossetti Can we remove the offset checking in gdrdrv_mmap()? To help users, we can also add a flag to gdr_map() such that gdr_map() automatically applies the offset (if any) before returning ptr_va.

drossetti:
@e-ago what if the user of gdrcopy were instead in charge of aligning the start address to the proper page boundary, and to properly handle the offset ?

drossetti:
I don't think we are ready to attack this problem, so removing 2.1 tag

PR #86, opened 2019-09-09
Ensure aligned allocations in sanity and copybw

pakmarkthub:
Problems:
- Issue #52.
- No unit test for unaligned mapping.
- Allocations in sanity and copybw may be unaligned because cuMemAlloc does not guarantee alignment.

This PR:
- adds a unit test in sanity to test unaligned mapping.
- creates alignedCUMemAlloc and alignedCUMemFree to handle alignment in copybw and sanity.

### DIFF
--- a/tests/common.hpp
+++ b/tests/common.hpp
@@ -25,132 +25,141 @@
 #include <map>
 #include <cuda.h>
 
-static std::map<CUdeviceptr, CUdeviceptr> _allocations;
+namespace gdrcopy::test {
+    static std::map<CUdeviceptr, CUdeviceptr> _allocations;
 
-static inline CUresult alignedCUMemAlloc(CUdeviceptr *pptr, size_t psize, bool set_sync_memops)
-{
-    CUresult ret = CUDA_SUCCESS;
-    CUdeviceptr ptr;
-    size_t size = psize + GPU_PAGE_SIZE - 1;
+    static inline CUresult gpuMemAlloc(CUdeviceptr *pptr, size_t psize, bool align_to_gpu_page = true, bool set_sync_memops = true)
+    {
+        CUresult ret = CUDA_SUCCESS;
+        CUdeviceptr ptr;
+        size_t size;
 
-    ret = cuMemAlloc(&ptr, size);
-    if (ret != CUDA_SUCCESS)
-        return ret;
+        if (align_to_gpu_page)
+            size = psize + GPU_PAGE_SIZE - 1;
+        else
+            size = psize;
 
-    if (set_sync_memops) {
-        unsigned int flag = 1;
-        ret = cuPointerSetAttribute(&flag, CU_POINTER_ATTRIBUTE_SYNC_MEMOPS, ptr);
-        if (ret != CUDA_SUCCESS) {
-            cuMemFree(ptr);
+        ret = cuMemAlloc(&ptr, size);
+        if (ret != CUDA_SUCCESS)
+            return ret;
+
+        if (set_sync_memops) {
+            unsigned int flag = 1;
+            ret = cuPointerSetAttribute(&flag, CU_POINTER_ATTRIBUTE_SYNC_MEMOPS, ptr);
+            if (ret != CUDA_SUCCESS) {
+                cuMemFree(ptr);
+                return ret;
+            }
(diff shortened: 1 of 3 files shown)


tests/common.hpp
```diff
@@ -22,6 +22,57 @@
...
+#include <map>
+#include <cuda.h>
+
+static std::map<CUdeviceptr, CUdeviceptr> _allocations;
+
+static inline CUresult alignedCUMemAlloc(CUdeviceptr *pptr, size_t psize, bool set_sync_memops)
+{
+    CUresult ret = CUDA_SUCCESS;
+    CUdeviceptr ptr;
+    size_t size = psize + GPU_PAGE_SIZE;
```
drossetti: PAGE_SIZE-1 should be enough

pakmarkthub:
@drossetti I fixed the code. Please re-review.

tests/common.hpp
```diff
@@ -22,6 +22,57 @@
...
+        unsigned int flag = 1;
+        ret = cuPointerSetAttribute(&flag, CU_POINTER_ATTRIBUTE_SYNC_MEMOPS, ptr);
+        if (ret != CUDA_SUCCESS) {
+            cuMemFree(ptr);
+            return ret;
+        }
+    }
+
+    if (ptr & GPU_PAGE_MASK != ptr)
+        *pptr = (ptr + GPU_PAGE_SIZE) & GPU_PAGE_MASK;
```
drossetti: you'll need -1 here as well

tests/sanity.cpp
```diff
@@ -180,16 +180,7 @@ START_TEST(basic)
...
-    ASSERTDRV(cuMemAlloc(&d_A, size));
-
-    unsigned int flag = 1;
-    ASSERTDRV(cuPointerSetAttribute(&flag, CU_POINTER_ATTRIBUTE_SYNC_MEMOPS, d_A));
-
-    uint32_t *init_buf = new uint32_t[size];
-    uint32_t *copy_buf = new uint32_t[size];
-
-    init_hbuf_walking_bit(init_buf, size);
-    memset(copy_buf, 0, sizeof(*copy_buf) * sizeof(uint32_t));
```
drossetti: this buffer initialization code has been removed, was it useless?

tests/common.hpp
```diff
@@ -22,6 +22,57 @@
 
 #pragma once
 
+#include <map>
+#include <cuda.h>
+
+static std::map<CUdeviceptr, CUdeviceptr> _allocations;
+
+static inline CUresult alignedCUMemAlloc(CUdeviceptr *pptr, size_t psize, bool set_sync_memops)
```
drossetti: style note, this is C++ so how about having:
- a new API which support both regular and aligned allocations
- possibly in a C++ namespace, like gdrcopy::
- with a nifty though overloaded name, like gpuMemAlloc() :)

tests/sanity.cpp
pakmarkthub: This "basic" unit test is for testing gdr_pin_buffer -- no mapping, no data copied. You can see that copy_buf and init_buf were not used anywhere.

tests/common.hpp
pakmarkthub: Can you clarify a few points?
1. Are you proposing that we should introduce a new API to libgdrapi? Or would we confine this in unit tests only?
2. Regular and aligned allocations would use the same interface, i.e., gpuMemAlloc(CUdeviceptr *, size_t) -- that bool set_sync_memops is optional. Overloading will not work. Maybe we should go with flags instead? We can make that set_sync_memops as another flag.

drossetti: just an idea, gpuMemAlloc(CUdeviceptr *pptr, size_t psize, bool set_sync_memops = true, bool align_to_gpu_page = true).
that would still be part of the internal test APIs

drossetti: probably wrapped in either anonymous or gdrcopy::test namespace

tests/sanity.cpp
drossetti: thanks for cleaning it up

pakmarkthub:
Fixed the code. @drossetti Please re-review.

PR #96, opened 2019-11-01
Extend the test timeout in sanity

pakmarkthub:
Problem:
- Some unit tests in sanity occasionally experience timeout on some systems.
- See nvbug 200559004 for more details.

This PR:
- sets the timeout of basic and data validation test cases to 60 seconds.
- sets the timeout of invalidation test case to 180 seconds.


pakmarkthub:
CUDA initialization takes a few seconds with GPU Driver v400 on some PPC64LE machines. The default timeout is 4 seconds. Hence, we experience the timeout, which is not related to the test. GDRCopy is not supported on that driver version and I have never seen the timeout when using supported driver versions. So, we might not need this PR.

PR #106, opened 2020-01-23
introduce CHECK_INC/LIBS in tests/Makefile

drossetti:
mainly to deal with cases when libbcheck is installed in a non-system dir

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,8 +1,8 @@
 DESTBIN ?= 
 CUDA ?= /usr/local/cuda
 
-CHECK_INC  ?= $(shell pkg-config --cflags-only-I check)
-CHECK_LIBS ?= $(shell pkg-config --libs check)
+CHECK_INC  := $(shell pkg-config --cflags-only-I check)
+CHECK_LIBS := $(shell pkg-config --libs check)
 
 GDRAPI_INC := ../include
 GDRAPI_SRC := ../src
@@ -10,7 +10,7 @@ GDRAPI_SRC := ../src
 CUDA_LIB := -L $(CUDA)/lib64 -L $(CUDA)/lib -L /usr/lib64/nvidia -L /usr/lib/nvidia
 CUDA_INC += -I $(CUDA)/include
 
-CPPFLAGS := $(CUDA_INC) -I $(GDRAPI_INC) -I $(GDRAPI_SRC) -I $(CUDA)/include -I $(CHECK_INC)
+CPPFLAGS := $(CUDA_INC) -I $(GDRAPI_INC) -I $(GDRAPI_SRC) -I $(CUDA)/include $(CHECK_INC)
 LDFLAGS  := $(CUDA_LIB) -L $(CUDA)/lib64 -L $(GDRAPI_SRC)
 COMMONCFLAGS := -O2
 CFLAGS   += $(COMMONCFLAGS)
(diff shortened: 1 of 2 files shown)


pakmarkthub:
libcheck comes with check.pc. Users can use `PKG_CONFIG_PATH=<folder-that-has-check.pc> make`. That would do the trick when they want to use libcheck installed in a custom location.

drossetti:
Interesting so it is simply a matter of documenting the PKG_CONFIG_PATH=<folder-that-has-check.pc> trick.
thanks

drossetti:
@pakmarkthub still wouldn't we need the INCLUDE part of my patch?

tests/Makefile
```diff
@@ -1,13 +1,16 @@
...
+CHECK_LIBS ?= $(shell pkg-config --libs check)
+
 GDRAPI_INC := ../include
 GDRAPI_SRC := ../src
 
 CUDA_LIB := -L $(CUDA)/lib64 -L $(CUDA)/lib -L /usr/lib64/nvidia -L /usr/lib/nvidia
 CUDA_INC += -I $(CUDA)/include
 
-CPPFLAGS := $(CUDA_INC) -I $(GDRAPI_INC) -I $(GDRAPI_SRC) -I $(CUDA)/include
+CPPFLAGS := $(CUDA_INC) -I $(GDRAPI_INC) -I $(GDRAPI_SRC) -I $(CUDA)/include -I $(CHECK_INC)
```
pakmarkthub: pkg-config --cflags-only-I returns "-I". So, no need for the leading "-I".

tests/Makefile
```diff
@@ -1,13 +1,16 @@
 DESTBIN ?= 
 CUDA ?= /usr/local/cuda
 
+CHECK_INC  ?= $(shell pkg-config --cflags-only-I check)
```
pakmarkthub: To help users who are not familiar with pkg-config, I would prefer to have CHECK_PATH possibly set by the users. If we detect CHECK_PATH, we use it to compose CHECK_INC and CHECK_LIBS. If not, get the flags from pkg-config.

drossetti: understood. I'll convert these assignments  into :=

drossetti:
see #118 for a related problem

drossetti:
@pakmarkthub I'd appreciate if you could take a look

pakmarkthub:
@drossetti 

LGTM.

Two small comments:

1. In README, could you also mention about using PKG_CONFIG_PATH with the packaging scripts such as `$ CUDA=<cuda-install-top-dir> PKG_CONFIG_PATH=<...> ./build-deb-packages.sh`.

2. While you are at it, can you make `build-rpm-packages.sh` support `PKG_CONFIG_PATH` as well?

drossetti:
> 1. In README, could you also mention about using PKG_CONFIG_PATH with the packaging scripts such as `$ CUDA=<cuda-install-top-dir> PKG_CONFIG_PATH=<...> ./build-deb-packages.sh`.

I'm not sure that would be enough, PKG_CONFIG_PATH would need to be passed down to rpmbuild somehow, most probably by extending the spec file.

> 2. While you are at it, can you make `build-rpm-packages.sh` support `PKG_CONFIG_PATH` as well?

Same for this case, I'm not convinced it is so easy.

And in the end, if the user is going to build the packages, because the dependency on libcheck is in their meta-data, PKG_CONFIG_PATH would also need to be complemented by forcing the build dependency first, and maybe the installation dependency later.

So I'd rather take more time to think this through.

Issue #108, opened 2020-02-21
copylat is not in generated rpms


davidcarver:
copylat is not in generated rpms

./tests/Makefile  and ./packages/gdrcopy.spec need to include copylat

drossetti:
@davidcarver thanks for reporting this issue.

cponder:
I don't see it in the source distribution, either:

      https://github.com/NVIDIA/gdrcopy/archive/v2.0.tar.gz

has these files in the tests subdirectory:

      Makefile  common.hpp  copybw.cpp  sanity.cpp

drossetti:
@cponder thank you
@davidcarver the problem should have been addressed in the master branch. So either pick that one or wait for the next minor release.

Issue #100, opened 2019-12-03
suggest requirements for package building in README


drossetti:
for example, for Debian:
```
sudo apt install build-essential devscripts debhelper
```

PR #110, opened 2020-03-01
Added copylat to the rpm and deb packages

pakmarkthub:
Problems:
- copylat is neither installed nor included in the deb / rpm packages.
- document dependencies for package building 

This PR:
- fixes issue #108.
- includes copylat in the deb package.
- modifies the _install_ command in tests/Makefile to also install copylat.
- modifies packages/build-deb-package.sh to respect user-defined PKG_CONFIG_PATH.
- introduces debian version in the deb packages.
- fixes issue #100.

### DIFF
--- a/packages/build-deb-packages.sh
+++ b/packages/build-deb-packages.sh
@@ -63,7 +63,7 @@ ex mv gdrcopy gdrcopy-${VERSION}
 ex tar czvf gdrcopy_${VERSION}.orig.tar.gz gdrcopy-${VERSION}
 
 ex cd $tmpdir/gdrcopy-${VERSION}
-ex debuild --set-envvar=CUDA=$CUDA -us -uc
+ex debuild --set-envvar=CUDA=${CUDA} --set-envvar=PKG_CONFIG_PATH=${PKG_CONFIG_PATH} -us -uc
 
 echo
 echo "Building dkms module ..."
--- a/packages/build-deb-packages.sh
+++ b/packages/build-deb-packages.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# Restart this number at 1 if MAJOR_VERSION or MINOR_VERSION changes
+# See https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
+DEBIAN_VERSION=2
+
 SCRIPT_DIR_PATH="$(cd "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")" && pwd)"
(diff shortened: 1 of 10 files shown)


packages/gdrcopy.spec
```diff
@@ -98,6 +98,7 @@ rm -rf $RPM_BUILD_DIR/%{name}-%{version}
 
 %files
 %{_prefix}/bin/copybw
+%{_prefix}/bin/copylat
```
drossetti: for updates to the spec, it is common practice to also bump the _release and to add a comment in the changelog

drossetti: similarly in debian/changelog ?

drossetti:
could you please also address #100 in this same PR?

pakmarkthub:
Fixed the PR as suggested.
@drossetti Please re-review.

packages/build-deb-packages.sh
```diff
@@ -56,26 +61,26 @@ ex cp README.md $tmpdir/gdrcopy/debian/README.source
...
-ex find . -type f -exec sed -i "s/@VERSION@/${VERSION}/g" {} +
+ex find . -type f -exec sed -i "s/@VERSION@/${FULL_VERSION}/g" {} +
 
 ex cd $tmpdir
 ex mv gdrcopy gdrcopy-${VERSION}
 ex tar czvf gdrcopy_${VERSION}.orig.tar.gz gdrcopy-${VERSION}
 
 ex cd $tmpdir/gdrcopy-${VERSION}
-ex debuild --set-envvar=CUDA=$CUDA -us -uc
+ex debuild --set-envvar=CUDA=${CUDA} --set-envvar=PKG_CONFIG_PATH=${PKG_CONFIG_PATH} -us -uc
```
drossetti: does this works even when PKG_CONFG_PATH is not set by the user?

packages/gdrcopy.spec
drossetti: please update chengelog section here too

packages/build-deb-packages.sh
pakmarkthub: Yes, it works fine. To be sure, can you test it on gc11 as well? The script should run successfully without setting PKG_CONFIG_PATH.

pakmarkthub:
@drossetti I have fixed the code. Please take a look.

packages/build-deb-packages.sh
drossetti: it builds ok there

PR #111, opened 2020-03-03
rename gdrapi internal functions

drossetti:
get rid of _bar suffix
- for internal functions
- in tracing messages


drossetti:
@pakmarkthub mind having a look?

PR #112, opened 2020-03-03
improve latency test

drossetti:
- print estimated bw, useful for large buffer sizes
- add -d <gpuid> param
- add warmup extra iterations and -w param


drossetti:
@pakmarkthub mind having a look?

tests/copylat.cpp
```diff
@@ -40,6 +40,8 @@ using namespace gdrcopy::test;
 // manually tuned...
 int num_write_iters = 10000;
 int num_read_iters = 100;
+int small_size_iter_factor = 1000;
```
pakmarkthub, line 43: I understand the intention and usefulness for small sizes. However, it changes what the number of iterations users specify. Is there a better way to do this or could you provide an explanation message? Currently, the users need to read the code in order to know that small sizes and large sizes use different number of iterations.

tests/copylat.cpp
```diff
@@ -49,10 +51,11 @@ int main(int argc, char *argv[])
     bool do_cumemcpy = false;
     struct timespec beg, end;
     double lat_us;
+    double bw;
```
pakmarkthub, line 54: Isn’t this redundant with copybw?

If you want to do shmoo for bw, is it better to rename the test? “copylat” doesn’t sound right anymore in that case.

PR #121, opened 2020-05-18
unblock multi-threaded use cases

### LINKED ISSUE #120: multi-threaded use cases lead to errno 13
for example
ERR:  ioctl error (errno=13)
[1589473761.717189] [xxxxxx:34848:1]    gdr_copy_md.c:122  UCX  ERROR gdr_pin_buffer failed. length :65536 ret:13

drossetti:
why: in gdrdrv, the current check for "same process" is too strong, and excludes tasks in same tgid.
how: relax the check on tsks

fixes #120

### DIFF
--- a/src/gdrdrv/gdrdrv.c
+++ b/src/gdrdrv/gdrdrv.c
@@ -266,7 +266,7 @@ struct gdr_info {
     // Pointer to the pid struct of the creator process. We do not use
     // numerical pid here to avoid issues from pid reuse.
     struct pid             *pid;
-    pid_t                   tgid;
+    struct pid             *tgid;
 
     // Address space unique to this opened file. We need to create a new one
     // because filp->f_mapping usually points to inode->i_mapping.
@@ -287,10 +287,10 @@ static int gdrdrv_check_same_process(gdr_info_t *info, struct task_struct *tsk)
     BUG_ON(0 == info);
     BUG_ON(0 == tsk);
     same_proc = (info->pid == task_pid(tsk))        // either exactly the same task
-        || (info->tgid == task_tgid_nr(tsk)) ; // or tasks belonging to the task group
+        || (info->tgid == task_tgid(tsk)) ; // or tasks belonging to the task group
     if (!same_proc) {
-        gdr_dbg("check failed, info:{pid=%p tgid=0x%x} this tsk={pid=%p tgid=0x%x}\n",
-                info->pid, info->tgid, task_pid(tsk), task_tgid_nr(tsk));
+        gdr_dbg("check failed, info:{pid=%p tgid=%p} this tsk={pid=%p tgid=%p}\n",
+                info->pid, info->tgid, task_pid(tsk), task_tgid(tsk));
     }
     return same_proc;
 }
@@ -324,7 +324,7 @@ static int gdrdrv_open(struct inode *inode, struct file *filp)
     // here we track the process owning the driver fd and prevent other process
     // to use it.
     info->pid = task_pid(current);
-    info->tgid = task_tgid_nr(current);
+    info->tgid = task_tgid(current);
 
     address_space_init_once(&info->mapping);
     info->mapping.host = inode;
--- a/src/gdrdrv/gdrdrv.c
+++ b/src/gdrdrv/gdrdrv.c
@@ -36,6 +36,9 @@
 #include <linux/timex.h>
 #include <linux/timer.h>
 #include <linux/sched.h>
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,11,0)
+#include <linux/sched/signal.h>
+#endif
 
 //-----------------------------------------------------------------------------
 
--- a/src/gdrdrv/gdrdrv.c
+++ b/src/gdrdrv/gdrdrv.c
@@ -266,6 +266,7 @@ struct gdr_info {
     // Pointer to the pid struct of the creator process. We do not use
     // numerical pid here to avoid issues from pid reuse.
     struct pid             *pid;
+    pid_t                   tgid;
 
     // Address space unique to this opened file. We need to create a new one
     // because filp->f_mapping usually points to inode->i_mapping.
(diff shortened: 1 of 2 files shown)


tests/sanity.cpp
```diff
@@ -1508,6 +1508,102 @@ BEGIN_GDRCOPY_TEST(invalidation_fork_child_gdr_pin_parent_with_tokens)
...
+    pthread_barrier_t barrier;
+};
+
+void *thr_fun_setup(void *data)
+{
+    mt_test_info *pt = static_cast<mt_test_info*>(data);
+    ASSERT(pt);
+    ASSERT_EQ(gdr_pin_buffer(pt->g, pt->d_buf, pt->size, 0, 0, &pt->mh), 0);
+    ASSERT_NEQ(pt->mh, null_mh);
+    ASSERT_EQ(gdr_map(pt->g, pt->mh, &pt->mapped_d_buf, pt->size), 0);
```
Akshay-Venkatesh, line 1527: I guess we're not breaking the test down further where threadA calls pin_buffer but a different thread calls map. This should probably be guaranteed by calling thread by grabbing lock before pin and releasing lock after map succeeds or fails.

drossetti: if you think it might be useful, I can add a finer grained test

Akshay-Venkatesh: From UCT viewpoint, [pin](https://github.com/openucx/ucx/blob/master/src/uct/cuda/gdr_copy/gdr_copy_md.c#L120) and [map](https://github.com/openucx/ucx/blob/master/src/uct/cuda/gdr_copy/gdr_copy_md.c#L126) are part of the same call [mem_reg](https://github.com/openucx/ucx/blob/master/src/uct/cuda/gdr_copy/gdr_copy_md.c#L106), so we're probably safe and we will likely add locks to ensure safe use of memory handle returned by gdr_map. If some other user of gdrcopy decides not to do this, then is it unsafe? For instance, what happens if threadA calls gdr_pin_buffer but the same memory handle is used by threadB and threadC concurrently?

drossetti: > we will likely add locks to ensure safe use of memory handle returned by gdr_map

why do you need locks?

>  For instance, what happens if threadA calls gdr_pin_buffer but the same memory handle is used by threadB and threadC concurrently?

for correctness, threadB and C would still need to wait for mem handle creation and mapping, right?
once the mem handle gets created and mapped, both B and C can use it, since it is a plain piece of virtual address space.

drossetti:
@pakmarkthub feel free to add your comments here

tests/sanity.cpp
Akshay-Venkatesh: > why do you need locks?

The aim was to ensure that two threads don't try to pin the same underlying cuda region concurrently. Is that considered safe? If so, then we don't have to use locks.

> for correctness, threadB and C would still need to wait for mem handle creation and mapping, right?

Agreed that there is a dependency on completion of pin_buffer.

Akshay-Venkatesh: > why do you need locks?

Just had a thought about this. Should we add the case where main thread allocates gpu mem and two child threads try to run pin+map on the same allocation?

Something like this:
```C++
    {
        pthread_t tid[2];
        ASSERT_EQ(pthread_barrier_init(&t.barrier, NULL, 2), 0);
        t.use_barrier = true;
        print_dbg("spawning two children threads\n");
        ASSERT_EQ(pthread_create(&tid[0], NULL, thr_fun_combined, &t), 0);
        ASSERT_EQ(pthread_create(&tid[1], NULL, thr_fun_combined, &t), 0);
        ASSERT_EQ(pthread_join(tid[0], NULL), 0);
        ASSERT_EQ(pthread_join(tid[1], NULL), 0);
    }
```

drossetti: > Just had a thought about this. Should we add the case where main thread allocates gpu mem and two child threads try to run pin+map on the same allocation?

Good idea. Just pushed that subtest. PLMK

Akshay-Venkatesh: Confirmed that the unit test and ucx-test still pass with the new changes. This confirms that it's not necessary to use locks to safely use pin buffer and map calls. 

Out of curiosity, I also checked to see if it's safe to call pin_buffer and map with different buffers and that also seems safe.

src/gdrdrv/gdrdrv.c
```diff
@@ -280,6 +281,20 @@ struct gdr_info {
 };
 typedef struct gdr_info gdr_info_t;
 
+static int gdrdrv_check_same_process(gdr_info_t *info, struct task_struct *tsk)
+{
+    int same_proc;
+    BUG_ON(0 == info);
+    BUG_ON(0 == tsk);
+    same_proc = (info->pid == task_pid(tsk))        // either exactly the same task
```
pakmarkthub, line 289: Do you still need to check task_pid? Doesn't the task_tgid checking alone already cover the task_pid checking?

tests/sanity.cpp
```diff
@@ -1547,6 +1657,7 @@ int main(int argc, char *argv[])
     tcase_add_test(tc_basic, basic);
     tcase_add_test(tc_basic, basic_with_tokens);
     tcase_add_test(tc_basic, basic_unaligned_mapping);
+    tcase_add_test(tc_basic, child_thread_pins_buffer);
```
pakmarkthub, line 1660: Would be good if you use the same naming scheme -- i.e., basic_*

src/gdrdrv/gdrdrv.c
drossetti: I had the same idea, but somehow wanted to be on the safe side. I'll get rid of the redundant tracking and check.

drossetti: done

tests/sanity.cpp
drossetti: good idea, done.

drossetti:
@pakmarkthub could you rereview ? thanks

nysal:
@drossetti This does not build on RHEL8 and I think newer kernel versions (I have only tried RHEL8). task_tgid was moved to linux/sched/signal.h

drossetti:
> @drossetti This does not build on RHEL8 and I think newer kernel versions (I have only tried RHEL8). task_tgid was moved to linux/sched/signal.h

@nysal that you for checking. Is there a backward compatible way of including signal.h, other than ifdef the include ?

drossetti:
> > @drossetti This does not build on RHEL8 and I think newer kernel versions (I have only tried RHEL8). task_tgid was moved to linux/sched/signal.h
> 
> @nysal that you for checking. Is there a backward compatible way of including signal.h, other than ifdef the include ?

it looks like linux/sched/signal.h has been introduced in [4.10-rc1](https://elixir.bootlin.com/linux/v4.11-rc1/source/net/unix/af_unix.c#L88).

nysal:
Thanks. The fix looks good to me.

Issue #131, opened 2020-08-03
dkms source not in deb (master)


NHDLarry:
Was just fooling with the master branch and found it isn't building the deb with sources in it. 

```
dpkg -i gdrdrv-dkms_2.0-2_amd64.deb
Preparing to unpack gdrdrv-dkms_2.0-2_amd64.deb ...
Unpacking gdrdrv-dkms:amd64 (2.0-2) over (2.0-2) ...
Setting up gdrdrv-dkms:amd64 (2.0-2) ...
Building for 4.15.0-112-generic
Building for architecture x86_64
This package appears to be a binaries-only package
 you will not be able to build against kernel 4.15.0-112-generic
 since the package source was not provided

```

drossetti:
It is confirmed and it is not a regression, e.g.

```
$ git checkout v2.0
...
$ CUDA=/usr/local/cuda-X.Y bash packages/build-deb-packages.sh
...
make[1]: Leaving directory '/tmp/gdr.EzKaQu/gdrdrv-dkms-2.0'
dh_installinit --name gdrdrv
dh_testdir
dh_testroot
dh_link
dh_strip
dh_compress
dh_fixperms
dh_installdeb
dh_shlibdeps
dh_gencontrol
dh_md5sums
dh_builddeb
dpkg-deb: building package 'gdrdrv-dkms' in '../gdrdrv-dkms_2.0_amd64.deb'.
 dpkg-genbuildinfo --build=binary
 dpkg-genchanges --build=binary >../gdrdrv-dkms_2.0_amd64.changes
dpkg-genchanges: info: binary-only upload (no source code included)
 dpkg-source --after-build gdrdrv-dkms-2.0
dpkg-buildpackage: info: binary-only upload (no source included)
```

drossetti:
branch fix_131 is an incomplete fix

drossetti:
> It is confirmed and it is not a regression, e.g.

my bad. v2.0 dkms package had the source embedded so it's a regression

PR #133, opened 2020-08-11
Fix the no-source bug in gdrdrv-dkms.deb (issue #131)

pakmarkthub:
Problems:
- See issue #131. 
- The dkms build scripts incorrectly handle DEBIAN\_VERSION (root cause).
- In build-deb-packages.sh, dpkg-buildpackage incorrectly marks dkms as binary-only package. This is the root cause of issue #131 but should be fixed nonetheless.

This PR:
- fixes issue #131.
- modifies build-deb-packages.sh and internal scripts to correctly handle DEBIAN\_VERSION.
- upgrades dkms source version to 3.0.

Pre-submit testing:
- On Ubuntu18.04.
- Tested install and purge for both gdrcopy.deb and gdrdrv-dkms.deb.
- Tested upgrade gdrdrv-dkms from version 2.0 to 2.1.


packages/build-deb-packages.sh
```diff
@@ -41,56 +41,60 @@ if [ "X$VERSION" == "X" ]; then
...
 
 echo
 echo "Building dkms module ..."
-ex cd $tmpdir/gdrcopy-${VERSION}/src/gdrdrv
+ex cd ${tmpdir}/gdrcopy-${VERSION}/src/gdrdrv
 ex make clean
 
-ex mkdir -p $tmpdir/gdrdrv-dkms-${VERSION}/
-ex cp -r $tmpdir/gdrcopy-${VERSION}/src/gdrdrv $tmpdir/gdrdrv-dkms-${VERSION}/gdrdrv-${VERSION}
-ex cp ${SCRIPT_DIR_PATH}/dkms.conf $tmpdir/gdrdrv-dkms-${VERSION}/gdrdrv-${VERSION}/
```
drossetti, line 80: I argue that the driver source code must not change when simply changing DEBIAN_VERSION, hence the dir should only versioned with MAJOR.MINOR

packages/build-deb-packages.sh
```diff
@@ -41,56 +41,60 @@ if [ "X$VERSION" == "X" ]; then
...
 fi
 
 echo "Building gdrcopy debian packages version ${FULL_VERSION} ..."
 
-echo "Working in $tmpdir ..."
+echo "Working in ${tmpdir} ..."
 
 ex cd ${TOP_DIR_PATH}
 
-ex mkdir -p $tmpdir/gdrcopy
```
drossetti, line 56: can you separate the ${tmpdir} cleanup in a separate patch within this same PR ?
it makes review much easier

pakmarkthub:
@drossetti 

This should fix as you suggested. Please re-review.

By the way, since you plan to create a different PR, I didn't add gdrcopy_ prefix to gdrdrv-dkms.deb here.

packages/build-deb-packages.sh
```diff
@@ -42,55 +42,65 @@ fi
...
-ex cd $tmpdir/gdrdrv-dkms-${VERSION}/
+dkmsdir="${tmpdir}/gdrdrv-dkms-${VERSION}"
+ex mkdir -p ${dkmsdir}
+ex cp -r ${tmpdir}/gdrcopy-${VERSION}/src/gdrdrv ${dkmsdir}/gdrdrv-${VERSION}
+ex cp ${SCRIPT_DIR_PATH}/dkms.conf ${dkmsdir}/gdrdrv-${VERSION}/
+ex cd ${dkmsdir}
 ex cp -r ${SCRIPT_DIR_PATH}/dkms/* .
-ex find . -type f -exec sed -i "s/@VERSION@/${FULL_VERSION}/g" {} +
+ex find . -type f -exec sed -i "s/@FULL_VERSION@/${FULL_VERSION}/g" {} +
+ex find . -type f -exec sed -i "s/@VERSION@/${VERSION}/g" {} +
```
drossetti, line 86: dkmsdir is a subdir of tmpdir, are these two 'find + sed' commands superfluous ?

drossetti:
> By the way, since you plan to create a different PR, I didn't add gdrcopy_ prefix to gdrdrv-dkms.deb here.

it makes sense!

packages/build-deb-packages.sh
pakmarkthub: No, they are not. These two commands are important in handling deb_version while keep the gdrdrv folder without deb_version.

Note that FULL_VERSION and VERSION refer to different strings.

drossetti: understood.
I referred to lines 64 and 65. But I see that you copy stuff into  ${dkmsdir}/gdrdrv-${VERSION} at line 81, which is later, so you need these extra sed.

PR #143, opened 2020-10-02
Add multiarch support for deb package

### LINKED ISSUE #142: support Ubuntu 20.04
On Ubuntu 20 DGX A100 system that has gdrcopy installed using apt install, gdrcopy libaries are in incorrect location and not found by the default library path

$ ll /usr/lib64
lrwxrwxrwx  1 root root    32 Apr 14 12:26 ld-linux-x86-64.so.2 -> /lib/x86_64-linux-gnu/ld-2.31.so*
lrwxrwxrwx  1 root root    14 Sep 16  2019 libgdrapi.so -> libgdrapi.so.2
lrwxrwxrwx  1 root root    16 Sep 16  2019 libgdrapi.so.2 -> libgdrapi.so.2.0
-rw-r--r--  1 root root 18600 Sep 16  2019 libgdrapi.so.2.0

This path is not used by Ubuntu and will not be searched. 

Libraries for Ubuntu should be here:
/usr/lib/x86_64-linux-gnu/ 

internally tracked as 3139008

flx42:
The previous Makefile was not following the GNU conventions:
https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

As a result, it was not possibly to properly override the library install in
the debian package script.

The deb package now correctly installs the library in the multiarch directory.
See https://wiki.debian.org/Multiarch/Implementation

Fix #142

Signed-off-by: Felix Abecassis <fabecassis@nvidia.com>

### DIFF
--- a/packages/gdrcopy.spec
+++ b/packages/gdrcopy.spec
@@ -58,7 +58,7 @@ echo "building"
 make -j CUDA=${CUDA}
 
 %install
-make install DESTDIR=$RPM_BUILD_ROOT PREFIX=%{_prefix}
+make install DESTDIR=$RPM_BUILD_ROOT prefix=%{_prefix} libdir=%{_libdir}
 make drv_install DESTDIR=$RPM_BUILD_ROOT 
 
 # Install gdrdrv service script
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,14 @@
-DESTDIR ?= 
-PREFIX ?= /usr
-DESTLIB ?= $(DESTDIR)/$(PREFIX)/lib64
-DESTBIN ?= $(DESTDIR)/$(PREFIX)/bin
+prefix      ?= /usr/local
+exec_prefix ?= $(prefix)
+libdir      ?= $(exec_prefix)/lib
+bindir      ?= $(exec_prefix)/bin
+includedir  ?= $(prefix)/include
+
+DESTDIR := $(abspath $(DESTDIR))
+DESTLIB = $(DESTDIR)$(libdir)
+DESTBIN = $(DESTDIR)$(bindir)
+DESTINC = $(DESTDIR)$(includedir)
+
 CUDA ?= /usr/local/cuda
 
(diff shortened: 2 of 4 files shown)


drossetti:
@pakmarkthub would you take a look?

packages/gdrcopy.spec
```diff
@@ -58,7 +58,7 @@ echo "building"
 make -j CUDA=${CUDA}
 
 %install
-make install DESTDIR=$RPM_BUILD_ROOT PREFIX=%{_prefix}
+make install DESTDIR=$RPM_BUILD_ROOT prefix=%{_prefix} libdir=/usr/lib64
```
pakmarkthub, line 61: Would be good to set `libdir=${_libdir}` so that it is consistent with the config under `%files` section.

pakmarkthub (approved):
One small comment. The rest looks good to me.

Before merging, I suggest that you as the QA to try installing v2.0 then upgrade to this version. We should check:
1. All files in v2.0 are cleaned up properly in every supported OS. As the installation path is changed, we should be careful that no garbage is left over.
2. The installation path is as intended in every OS. And possibly the path is in the search space with the out-of-the-box OS configuration.

flx42:
> All files in v2.0 are cleaned up properly in every supported OS. As the installation path is changed, we should be careful that no garbage is left over.

This is handled by the package manager, not the package itself. 

> The installation path is as intended in every OS. And possibly the path is in the search space with the out-of-the-box OS configuration.

I checked the deb and rpm packages. The deb package is multiarch enabled so it's fine for debian-based distros.
The rpm package installs to /usr/lib64, and AFAIK it is always the package install path for RPM based systems (RHEL, CentOS, Fedora, OpenSUSE).

packages/gdrcopy.spec
flx42: Fixed, thanks!

drossetti:
@flx42 thanks for looking into this

PR #149, opened 2020-11-05
Small fixes to RPM spec file

wpoely86:
- Use `%{CUDA}` instead of `${CUDA}`.
- Run depmod specifically for the kernel that the module was build for instead of the running kernel


drossetti:
@pakmarkthub please have a look

pakmarkthub:
Thank you for the fixes.

Issue #126, opened 2020-07-29
sanity fails to detect missing kernel driver


drossetti:
discovered by chance:
```
drossetti@dgx-1v-13:~/gdrcopy$ LD_LIBRARY_PATH=$PWD/src ./tests/sanity
Running suite(s): Sanity
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:181
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:218
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:276
Assertion "!((t.g) == ((void*)0))" failed at sanity.cpp:1583
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:358
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:471
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:547
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:626
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:772
Assertion "(read(read_fd, &cont, sizeof(int))) == (sizeof(int))" failed at sanity.cpp:740
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:881
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:1021
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:1134
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:1134
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:1264
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:1388
Assertion "!((g) == ((void*)0))" failed at sanity.cpp:1470
100%: Checks: 15, Failures: 0, Errors: 0
```

pakmarkthub:
The PRs has been merged. Closing the issue.

PR #155, opened 2020-12-07
Improve gdr_open error message in tests

pakmarkthub:
Problems:
- Issue #126.
- In tests (copybw, copylat, sanity), error from gdr_open is detected with assert.
- Most of the time, gdr_open error occurs because gdrdrv is not loaded. However, users most likely cannot guess from the assertion error message.

This PR:
- changes the error message to "gdr_open error: Is gdrdrv driver installed and loaded?". Applied to copybw, copylat, and sanity.

Presubmit testing:
- On gc02 with and without gdrdrv loaded.

### DIFF
--- a/tests/common.hpp
+++ b/tests/common.hpp
@@ -111,6 +111,15 @@ START_TEST(__testname) {                                                \
 #define BEGIN_CHECK do
 #define END_CHECK while(0)
 
+#define GDR_OPEN_SAFE() ({                                                          \
+        gdr_t _g = gdr_open();                                                      \
+        if (!_g) {                                                                  \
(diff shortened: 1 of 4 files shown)


tests/common.hpp
```diff
@@ -111,6 +111,15 @@ START_TEST(__testname) {                                                \
 #define BEGIN_CHECK do
 #define END_CHECK while(0)
 
+#define GDR_OPEN_SAFE() ({                                                              \
```
drossetti, line 114: I prefer inline functions over macros, unless strictly necessary to pass info regarding the original call site

drossetti (changes requested):

tests/common.hpp
pakmarkthub: Thank you Davide. Fixed.

Issue #156, opened 2020-12-08
sanity does not properly summarize failure


pakmarkthub:
The last line of `sanity` always reports `100%: Checks: 15, Failures: 0, Errors: 0` although it encounters errors.

pakmarkthub:
Closing as the PR has been merged.

Issue #137, opened 2020-09-11
improve output of sanity test


drossetti:
customer feedback: there is a lot of noise in the output, even when not run with logs enabled, and the user may miss the last line which shows 0 errors.
```
Running suite(s): Sanity 
ERR:  error Invalid argument(22) while mapping handle ce00a0, rounded_size=69632 offset=0 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  sse4_1=1 avx=1 sse=1 sse2=1 
DBG:  using AVX implementation of gdr_copy_to_bar 
DBG:  using SSE4_1 implementation of gdr_copy_from_bar 
warning: buffer size 327669 is not dword aligned, ignoring trailing bytes 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
DBG:  wc_mapping=1 
CUDA error: CUDA_ERROR_INVALID_DEVICE 
Assertion "CUDA_SUCCESS == result" failed at sanity.cpp:67 
Assertion "(read(read_fd, &child_data, sizeof(int))) == (sizeof(int))" failed at sanity.cpp:817 
DBG:  wc_mapping=1 
ERR:  error Permission denied(13) while mapping handle ce0570, rounded_size=65536 offset=0 
CUDA error: CUDA_ERROR_INVALID_DEVICE 
Assertion "CUDA_SUCCESS == result" failed at sanity.cpp:67 
CUDA error: CUDA_ERROR_INVALID_DEVICE 
Assertion "CUDA_SUCCESS == result" failed at sanity.cpp:67 
CUDA error: CUDA_ERROR_INVALID_DEVICE 
Assertion "CUDA_SUCCESS == result" failed at sanity.cpp:67 
Assertion "sendfd(pair[1], fd) >= 0" failed at sanity.cpp:1399 
ERR:  ioctl error (errno=22) 
100%: Checks: 15, Failures: 0, Errors: 0 
```

pakmarkthub:
There actually are failures but the last line reports incorrectly. The PR should fix it.

pakmarkthub:
The PR has been merged. Closing.

PR #157, opened 2020-12-09
Fix success/failure summarization in sanity

pakmarkthub:
Problems:
- Issue #156, #137, and #126.
- `sanity` always reports success in the summarization line (last line) even with failures.

This PR:
- fixes issue #156, #137, and #126.

Presubmit testing:
- On gc02 with and without gdrdrv installed (simulated success and failure).
- Check output and make sure that it is compatible with DVS format.

### DIFF
--- a/tests/common.hpp
+++ b/tests/common.hpp
@@ -77,8 +77,10 @@ START_TEST(__testname) {                                                \
             print_dbg("&&&& PASSED %s\n", gdrcopy::test::testname);     \
         else if (__child_exit_status == EXIT_WAIVED)                    \
             print_dbg("&&&& WAIVED %s\n", gdrcopy::test::testname);     \
-        else                                                            \
+        else {                                                          \
             print_dbg("&&&& FAILED %s\n", gdrcopy::test::testname);     \
+            ck_abort();                                                 \
+        }                                                               \
     }                                                                   \
 } END_TEST
 
--- a/tests/common.hpp
+++ b/tests/common.hpp
@@ -61,17 +61,17 @@ START_TEST(__testname) {                                                \
     if (__pid < 0) {                                                    \
         print_dbg("Cannot fork\n");                                     \
         print_dbg("&&&& FAILED " # __testname "\n");                    \
-        exit(EXIT_FAILURE);                                             \
+        ck_abort();                                                     \
     }                                                                   \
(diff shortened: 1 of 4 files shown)


tests/common.hpp
```diff
@@ -111,6 +113,15 @@ START_TEST(__testname) {                                                \
 #define BEGIN_CHECK do
 #define END_CHECK while(0)
 
+#define GDR_OPEN_SAFE() ({                                                              \
```
drossetti, line 116: can you use an inline function instead of a macro?

drossetti (approved):
except for using an inline function, LGTM

tests/common.hpp
```diff
@@ -111,6 +113,15 @@ START_TEST(__testname) {                                                \
...
+        gdr_t _g = gdr_open();                                                          \
+        if (!_g) {                                                                      \
+            fprintf(stderr, "gdr_open error: Is gdrdrv driver installed and loaded?\n");\
+            exit(EXIT_FAILURE);                                                         \
```
drossetti, line 120: should you not call ck_abort() ?

pakmarkthub: Fixed.

tests/common.hpp
pakmarkthub: Fixed.

tests/copybw.cpp
```diff
@@ -144,8 +144,7 @@ int main(int argc, char *argv[])
     ASSERT_NEQ(init_buf, (void*)0);
     init_hbuf_walking_bit(init_buf, size);
 
-    gdr_t g = gdr_open();
-    ASSERT_NEQ(g, (void*)0);
```
drossetti, line 148: it looks like you removed the ASSERT_NEQ(). It's not a big deal gdr_open_safe() already handles the error, but it breaks the pattern without a reason.
I suggest to either remove ASSERT_NEQ() from all gdr_open_safe() call sites, or to preserve the one here.

drossetti (approved):
small fix, lgtm otherwise

tests/copybw.cpp
pakmarkthub: Since ASSERT_NEQ loss its meaning after we have gdr_open_safe, I choose to remove it from all call sites.

Issue #169, opened 2021-01-30
prepare for 2.2 release


drossetti:
- update version to 2.2 everywhere
- snap 2.2-beta1 tag

pakmarkthub:
Merged. Closing...

PR #130, opened 2020-07-31
update meta-data for 2.1 release

drossetti:
@pakmarkthub please review

PR #171, opened 2021-02-01
Change version to 2.2-1 and update changelogs

pakmarkthub:
This PR:
- is for issue #169.
- bumps up version to 2.2. 
- resets package version to 1.
- updates copyright year.
- adds copyright text to all Makefiles, config_arch, build_*_packages.sh.
- updates changelogs.
- updates README to use lower-case `prefix` instead of the upper-case one.


packages/debian/changelog
```diff
@@ -1,5 +1,21 @@
 gdrcopy (@FULL_VERSION@) unstable; urgency=low
```
drossetti, line 1: remind me why we need unstable here

drossetti (changes requested):
using https://github.com/NVIDIA/gdrcopy/pull/130 as reference:
- update CHANGELOG.md
- reset DEBIAN_VERSION to 1 in build-deb-packages.sh

packages/debian/changelog
pakmarkthub: Changed to stable. It is a legacy I forgot to change. Thanks for pointing it out.

pakmarkthub:
@drossetti Can I ask you to take a look at the current revision?

drossetti (approved):

lukeyeager:
@pakmarkthub did you mean to add a tag for this commit, too? There isn't one, currently.
https://github.com/NVIDIA/gdrcopy/tags

pakmarkthub:
@lukeyeager Version 2.2 is not released yet. We are preparing the code for release in the near future.

Issue #165, opened 2021-01-21
copybw and copylat silently exit if the GPU does not support GPUDirect RDMA


pakmarkthub:
`BREAK_IF_NEQ(gdr_pin_buffer(g, d_A, size, 0, 0, &mh), 0);` is problematic. The test is waived but no message to notify the user. Also, the applications exit successfully. This needs to be fix.

PR #184, opened 2021-03-24
Print more meaningful error messages in copybw and copylat when gdr_pin_buffer fails

pakmarkthub:
Issue:
- See #165 and #189.
- copybw and copylat silently fail when running on GPU that doesn't support GPUDirect RDMA.

This PR:
- modifies copybw and copylat to print meaningful error messages when gdr_pin_buffer fails.

Presubmit testing:
- On GeForce RTX 2070

```
$ LD_LIBRARY_PATH=../src/:$LD_LIBRARY_PATH ./copylat
GPU id:0; name: GeForce RTX 2070; Bus id: 0000:01:00
selecting device 0
device ptr: 0x7f47ff000000
allocated size: 16777216

error in gdr_pin_buffer with code=22
Does the selected GPU support GPUDirect RDMA?
Assertion "(status) == (0)" failed at copylat.cpp:195
```

### DIFF
--- a/tests/copybw.cpp
+++ b/tests/copybw.cpp
@@ -42,8 +42,6 @@ int num_read_iters  = 100;
 
 int main(int argc, char *argv[])
 {
-    int status = 0;
-
     size_t _size = 128*1024;
     size_t copy_size = 0;
     size_t copy_offset = 0;
@@ -130,24 +128,6 @@ int main(int argc, char *argv[])
     cout << "selecting device " << dev_id << endl;
     ASSERTDRV(cuDeviceGet(&dev, dev_id));
 
-    #if CUDA_VERSION >= 11030
-    int drv_version;
-    ASSERTDRV(cuDriverGetVersion(&drv_version));
-
-    // Starting from CUDA 11.3, CUDA provides an ability to check GPUDirect RDMA support.
-    if (drv_version >= 11030) {
-        int gdr_support = 0;
-        ASSERTDRV(cuDeviceGetAttribute(&gdr_support, CU_DEVICE_ATTRIBUTE_GPU_DIRECT_RDMA_SUPPORTED, dev));
-
-        if (!gdr_support)
-            cerr << "This GPU does not support GPUDirect RDMA." << endl;
-
-        ASSERT_NEQ(gdr_support, 0);
-    }
-
-    // For older versions, we fall back to detect this support when calling gdr_pin_buffer.
-    #endif
-
 
     CUcontext dev_ctx;
     ASSERTDRV(cuDevicePrimaryCtxRetain(&dev_ctx, dev));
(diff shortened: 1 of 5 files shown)


tests/copybw.cpp
```diff
@@ -149,8 +151,12 @@ int main(int argc, char *argv[])
     gdr_mh_t mh;
     BEGIN_CHECK {
         // tokens are optional in CUDA 6.0
-        // wave out the test if GPUDirectRDMA is not enabled
-        BREAK_IF_NEQ(gdr_pin_buffer(g, d_A, size, 0, 0, &mh), 0);
+        status = gdr_pin_buffer(g, d_A, size, 0, 0, &mh);
+        if (status != 0) {
+            cerr << "error in gdr_pin_buffer with code=" << status << endl;
+            cerr << "Does the selected GPU support GPUDirect RDMA?" << endl;
```
drossetti, line 157: the second diagnostic msg is barely useful.
we could use CU_DEVICE_ATTRIBUTE_GPU_DIRECT_RDMA_SUPPORTED in CUDA 11.3 to distinguish the GDR supported cases from other cases

drossetti: just filed #189

tests/copylat.cpp
```diff
@@ -184,8 +186,12 @@ int main(int argc, char *argv[])
     gdr_mh_t mh;
     BEGIN_CHECK {
         // tokens are optional in CUDA 6.0
-        // wave out the test if GPUDirectRDMA is not enabled
-        BREAK_IF_NEQ(gdr_pin_buffer(g, d_A, size, 0, 0, &mh), 0);
+        status = gdr_pin_buffer(g, d_A, size, 0, 0, &mh);
+        if (status != 0) {
+            cerr << "error in gdr_pin_buffer with code=" << status << endl;
+            cerr << "Does the selected GPU support GPUDirect RDMA?" << endl;
```
drossetti, line 192: same as above

tests/copybw.cpp
drossetti: in practice, I'm asking to to simply print the first msg.

drossetti (changes requested):

tests/copybw.cpp
pakmarkthub: Done.

tests/copylat.cpp
pakmarkthub: Done.

tests/copybw.cpp
```diff
@@ -128,6 +130,25 @@ int main(int argc, char *argv[])
     cout << "selecting device " << dev_id << endl;
     ASSERTDRV(cuDeviceGet(&dev, dev_id));
 
+    #if CUDA_VERSION >= 11030
```
drossetti, line 133: can we refactor this code into an helper function used by all the tests?

drossetti (changes requested):
just one last thing

tests/copybw.cpp
pakmarkthub: Done. Can I ask you to re-review?

drossetti (approved):

Issue #188, opened 2021-04-05
gdrcopy deb dependency on gdrdrv-dkms is problematic for containers


cliffwoolley:
In a container userspace, we would typically install the gdrcopy_*.deb (gdrapi), but _not_ gdrdrv-dkms.  Unfortunately, gdrcopy_*.deb has a dependency on the kernel-mode driver deb package.  While that's fine on bare metal, it doesn't really make sense for a containerized deployment.

The way CUDA handles this is to have `cuda-toolkit` and `cuda-drivers` packages that do not advertise any dependency on each other, and there is a `cuda` metapackage that depends on both for ease of installation.  Could we do something like that here?

pakmarkthub:
The PR has been merged. Closing this issue.

Issue #187, opened 2021-04-05
Building deb package builds the tests, which adds dependency on CUDA when building debs


cliffwoolley:
When running `build-deb-packages.sh`, it ends up implicitly building the tests (both because `debian/rules` builds the `exes` and, even if you turn that off, because the build process installs the built debs at the end, and `make install` also leads to `exes_install` which again builds the tests.

This is unfortunate, because building the actual deb packages does not actually depend on the CUDA Toolkit being installed; only the tests depend on CUDA.  So for example if I hack up the Makefile as follows, then `build-deb-packages.sh` will run fine without CUDA installed:

```
--- Makefile.orig       2021-04-04 23:17:02.799050000 -0700
+++ Makefile.new        2021-04-04 23:18:11.437725000 -0700
@@ -59,10 +59,8 @@
        $(MAKE) LIB_MAJOR_VER=$(LIB_MAJOR_VER) LIB_MINOR_VER=$(LIB_MINOR_VER)

 exes: lib
-       cd tests && \
-       $(MAKE) CUDA=$(CUDA)

-install: lib_install exes_install
+install: lib_install

 lib_install: lib
        @ echo "installing in $(DESTLIB) $(DESTINC)..." && \
```

drossetti:
@cliffwoolley would moving the tests to a separate package solve the problem?

cliffwoolley:
As far as I can tell, the tests aren't packaged anyway.  It seemed like they were getting built more as a side-effect.

pakmarkthub:
@cliffwoolley Sorry for the delay. I just have time to focus on this project.

I rechecked and the tests are in the deb package (see below, you can see copybw, copylat, sanity). Are they not in the package you generated using the script?

Let's go back to your request. Is my understanding correct that you don't want to install CUDA in order to use the script? I don't think you want the CUDA dependency be removed from `gdrcopy.deb` because there is no such dependency check in the first place.

If my understanding is correct, I think I can provide an option in the `build-deb-packages.sh` script to **NOT** build the tests. In that case, we also need to put the tests into another package as Davide suggested.

```
$ dpkg -c gdrcopy_2.2-1_amd64.ubuntu20_04.deb 
drwxr-xr-x root/root         0 2021-02-02 03:59 ./
drwxr-xr-x root/root         0 2021-02-02 03:59 ./usr/
drwxr-xr-x root/root         0 2021-02-02 03:59 ./usr/bin/
[...]

PR #192, opened 2021-04-20
Remove gdrdrv-dkms dependency enforcement from gdrcopy main package

pakmarkthub:
Problem: See issue #188.

This PR:
- removes gdrdrv-dkms dependency enforcement from gdrcopy.deb.
- removes gdrcopy-kmod dependency enforcement from gdrcopy.rpm.
- adds libsubunit0 dependency enforcement to gdrcopy.deb. `sanity` requires it but the dependency is not declared in the deb package. For rpm, this dependency is auto-discovered.

With this PR, now the package metadata looks like
```
$ dpkg-deb -I gdrcopy_2.2-1_amd64.deb 
 new Debian package, version 2.0.
 size 54016 bytes: control archive=900 bytes.
     388 bytes,    12 lines      control              
     494 bytes,     8 lines      md5sums              
      20 bytes,     1 lines      shlibs               
      74 bytes,     2 lines      triggers             
 Package: gdrcopy
 Version: 2.2-1
 Architecture: amd64
 Maintainer: Davide Rossetti <drossetti@nvidia.com>
 Installed-Size: 176
 Depends: libsubunit0, libc6 (>= 2.14)
 Section: libs
 Priority: optional
 Multi-Arch: same
 Homepage: https://github.com/NVIDIA/gdrcopy
 Description: A low-latency GPU memory copy library
  A low-latency GPU memory copy library based on NVIDIA GPUDirect RDMA technology.
```

```
$ rpm -qp --requires gdrcopy-2.2-1.x86_64.rpm
check
libc.so.6()(64bit)
libc.so.6(GLIBC_2.11)(64bit)
libc.so.6(GLIBC_2.14)(64bit)
libc.so.6(GLIBC_2.2.5)(64bit)
libc.so.6(GLIBC_2.3)(64bit)
libc.so.6(GLIBC_2.3.4)(64bit)
libc.so.6(GLIBC_2.4)(64bit)
libgdrapi.so.2()(64bit)
libm.so.6()(64bit)
libm.so.6(GLIBC_2.2.5)(64bit)
libpthread.so.0()(64bit)
libpthread.so.0(GLIBC_2.2.5)(64bit)
libpthread.so.0(GLIBC_2.3.3)(64bit)
librt.so.1()(64bit)
librt.so.1(GLIBC_2.2.5)(64bit)
librt.so.1(GLIBC_2.3.3)(64bit)
libstdc++.so.6()(64bit)
libstdc++.so.6(GLIBCXX_3.4)(64bit)
libstdc++.so.6(GLIBCXX_3.4.11)(64bit)
libstdc++.so.6(GLIBCXX_3.4.9)(64bit)
libsubunit.so.0()(64bit)
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rtld(GNU_HASH)
```

### DIFF
--- a/packages/debian/control
+++ b/packages/debian/control
@@ -11,6 +11,6 @@ Homepage: https://github.com/NVIDIA/gdrcopy
 Package: gdrcopy
 Architecture: any
 Multi-Arch: same
-Depends: gdrdrv-dkms (>= @FULL_VERSION@), ${shlibs:Depends}, ${misc:Depends}
+Depends: ${shlibs:Depends}, ${misc:Depends}
 Description: A low-latency GPU memory copy library 
  A low-latency GPU memory copy library based on NVIDIA GPUDirect RDMA technology.
--- a/packages/debian/control
+++ b/packages/debian/control
@@ -11,6 +11,6 @@ Homepage: https://github.com/NVIDIA/gdrcopy
 Package: gdrcopy
 Architecture: any
 Multi-Arch: same
-Depends: ${shlibs:Depends}, ${misc:Depends}
+Depends: libsubunit0, ${shlibs:Depends}, ${misc:Depends}
 Description: A low-latency GPU memory copy library 
  A low-latency GPU memory copy library based on NVIDIA GPUDirect RDMA technology.
--- a/packages/gdrcopy.spec
+++ b/packages/gdrcopy.spec
@@ -53,7 +53,7 @@ License:        MIT
 URL:            https://github.com/NVIDIA/gdrcopy
 Source0:        %{name}-%{version}.tar.gz
 BuildRequires:  gcc kernel-headers check-devel
-Requires:       %{name}-%{kmod} check
+Requires:       check
 
 %package devel
 Summary: The development files


pakmarkthub:
Hi @cliffwoolley,

Can I ask you to take a look?

Note that this PR is for issue #188 only. I will create a different PR for issue #187.

cliffwoolley:
Is the idea for adding libsubunit dependency just that it's presently a silent requirement, and your change for https://github.com/NVIDIA/gdrcopy/issues/187 will split it back out again?

I realized in https://github.com/NVIDIA/gdrcopy/issues/187 that it might be better for backward consistency to have the gdrcopy package just become a metapackage and _not_ drop the gdrdrv-dkms dependency from the metapackage; any thoughts on that?  If you think it's best to drop it and are sure this won't pose problems for those installing one and assuming it will install the other for them because it always did, then ok, LGTM.  But I do think you should consider that extra layer of indirection.

pakmarkthub:
Yes, `libsubunit` is a silent requirement. So, I add it here to make thing correct.

In the current code (not this PR), we don't enforce libcuda as a dependency of gdrcopy package even though the tests need it. This is so that users can install CUDA via run file and gdrcopy from package. We can relax the dependency checking with gdrdrv as well. Users will get notify if they try to use libgdrapi without gdrdrv anyway.

PR for #187 will require more test. We have to ensure that users can upgrade from real package in the current version to metapackage in the next version. Or maybe we will use a different name for the metapackage to avoid compatibility issues.

cliffwoolley:
> In the current code (not this PR), we don't enforce libcuda as a dependency of gdrcopy package even though the tests need it.

This is fine with me and all; I was just thinking of upgrade compatibility for existing users.

drossetti (approved):

PR #193, opened 2021-04-21
Document BAR1 reuse issue to the Restrictions and known issues section

pakmarkthub:
This PR documents the BAR1 reuse issue in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,10 +291,10 @@ Neither CUDA Runtime nor Driver APIs guarantees that GPU memory allocation
 functions return aligned addresses. Users are responsible for proper alignment
 of addresses passed to the library.
 
-In some GPU driver versions, pinning the same GPU address multiple times consume
-additional BAR1 space. This is because the space is not properly reused. If you
-encounter this issue, we suggest that you try the latest version of NVIDIA GPU
-driver.
+In some GPU driver versions, pinning the same GPU address multiple times
+consumes additional BAR1 space. This is because the space is not properly
+reused. If you encounter this issue, we suggest that you try the latest version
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -291,6 +291,13 @@ Neither CUDA Runtime nor Driver APIs guarantees that GPU memory allocation
 functions return aligned addresses. Users are responsible for proper alignment
 of addresses passed to the library.
 
+Normally, pinning the same GPU address multiple times consumes the same amount
+of BAR1 space as doing it for the first time on that address because of the BAR1
+address reuse. However, GPU driver has a bug that prevents address reuse when
+pinning larger than a certain size (depending on platforms and GPU types). The
```
drossetti, line 297: I would be more generic, as in "some GPU drivers have a bug that ... "
That is because you never know on which branches the fix will be backported.

drossetti (changes requested):

README.md
pakmarkthub: Agreed. Fixed the paragraph.

README.md
```diff
@@ -291,6 +291,11 @@ Neither CUDA Runtime nor Driver APIs guarantees that GPU memory allocation
 functions return aligned addresses. Users are responsible for proper alignment
 of addresses passed to the library.
 
+In some GPU driver versions, pinning the same GPU address multiple times consume
```
drossetti, line 294: "consume" -> "consumes"

drossetti (changes requested):
small nit

README.md
pakmarkthub: Nice catch. Fixed.

Issue #190, opened 2021-04-09
Building RPMs without dkms


bne86:
Hi,

for us its very important that we are able to build the kernel modules without DKMS. DKMS is pulling in several dependencies we do not want to install on our HPC machine (9XX nodes, 4 A100 each, diskless) because it would use to much memory.

Actually, we do not build the kmods on the nodes, we have a build-system that create RPMs we at the end use on a repository to distribute it to our several systems.

With the latest gdrcopy 2.2 changes it looks like you started to only support DKMS.  Do you think its possible to support two specfiles, one with and one without DKMS? Otherwise we always have to reverse engineer your dkms specfiles.

best regards

pakmarkthub:
Hi @bne86,

We plan to release pre-built RPM and DEB packages in the near future. The DKMS for RPM build script is in preparation for that release.

Since you have a use case for kmod, we will bring back the kmod build script. However, kmod will not be a part of those pre-built packages -- so you need to build it yourself using the provided script. Is this ok with you?

Sincerely yours

bne86:
Definitely, thanks!

pakmarkthub:
The PR has been merged. Closing this issue.

PR #194, opened 2021-04-22
Bring back non-dkms kmod rpm

pakmarkthub:
Problem
- See issue #190.

This PR:
- brings back kmod rpm generator script.
- makes kmod rpm build optional (default is to not build).
- makes kmod rpm package name specific to nvidia driver version and linux kernel version.
- renames dkms-related names to make it more distinguishable from kmod.
- changes scripts in dkms rpm package to do discover kernel version at runtime instead of at build time. On the other hand, kmod rpm uses static kernel version because it is meant to be used with only that specific version.


packages/gdrcopy.spec
```diff
@@ -83,23 +103,40 @@ GDRCopy, a low-latency GPU memory copy library and a kernel-mode driver, built o
...
+make -j8 NVIDIA_SRC_DIR=%{NVIDIA_SRC_DIR} driver
+%endif
 
 %install
 # Install gdrcopy library and tests
 make install DESTDIR=$RPM_BUILD_ROOT prefix=%{_prefix} libdir=%{_libdir}
 
+%if %{BUILD_KMOD_NONDKMS} > 0
+# Install gdrdrv driver
+make drv_install DESTDIR=$RPM_BUILD_ROOT
```
lexming, line 130: Tested this PR and with the following change works well on our end. The `make drv_install` also has to define the `NVIDIA_SRC_DIR`
```suggestion
make drv_install NVIDIA_SRC_DIR=%{NVIDIA_SRC_DIR} DESTDIR=$RPM_BUILD_ROOT
```

pakmarkthub: Thank you for the suggestion. I don't think we need to add `NVIDIA_SRC_DIR=%{NVIDIA_SRC_DIR}` here because it is already done in the compilation phase (see line 121 of this PR). Have you encountered a problem and that line needs to be added here to fix it?

lexming: Yes, indeed, otherwise I would not have made the suggestion. Without this change, the `%install` step ignores the definition of `NVIDIA_SRC_DIR` done at the top of the spec file, causing the `make` command to rely on the outcome of `NVIDIA_SRC_DIR ?= $(shell find /usr/src/nvidia-* -name "nv-p2p.h"|head -1|xargs dirname || echo "NVIDIA_DRIVER_MISSING")` in  `src/gdrdrv/Makefile`. This is wrong for two reasons:
* the `%build` step and the `%install` will very likely uses different `NVIDIA_SRC_DIR`
* the `%install` step fails if files are not found in `/usr/src/nvidia-*`

pakmarkthub: I am trying to understand this problem since I cannot reproduce it.

The `%install` step should not try to rebuild gdrdrv again since it has already been done. To test this, I change `NVIDIA_SRC_DIR ?= "UNDEFINED"` inside `src/gdrdrv/Makefile`. So, it always fails if I don't provide NVIDIA_SRC_DIR when calling the packaging script. Then, I run `CUDA=/usr/local/cuda NVIDIA_SRC_DIR=/usr/src/nvidia-460.73.01/nvidia ./build-rpm-packages.sh -m`. Everything works well.

Let me confirm. Did you encounter the problem when trying to create the rpm packages or when you use the package?

lexming: The script `build-rpm-packages.sh` works because it sets `NVIDIA_SRC_DIR` on its own. This issue concerns the `.spec` file, which cannot be used _standalone_. For instance, directly building the RPM with  `rpmbuild` from the spec file fails (unless the nvidia sources are in `/usr/src/nvidia-*`)
```
$ rpmbuild -ba --define "CUDA /path/to/cuda" --define "GDR_VERSION 2.2" --define "KVERSION $(uname -r)" --define "MODULE_LOCATION kernel/drivers/misc/" --define "NVIDIA_DRIVER_VERSION 460.73.01" --define "NVIDIA_SRC_DIR $HOME/nvidia-kmod-460.73.01-x86_64/kernel/nvidia" --define "BUILD_KMOD_NONDKMS 1" packages/gdrcopy.spec
[...]
cd src/gdrdrv && \
make 
find: '/usr/src/nvidia-*': No such file or directory
dirname: missing operand
Try 'dirname --help' for more information.
make[1]: Entering directory `/home/lexming/rpmbuild/BUILD/gdrcopy-2.2/src/gdrdrv'
Picking NVIDIA driver sources from NVIDIA_SRC_DIR=NVIDIA_DRIVER_MISSING. If that does not meet your expectation, you might have a stale driver still around and that might cause problems.
make[2]: Entering directory `/usr/src/kernels/3.10.0-1160.24.1.el7.x86_64'
find: '/usr/src/nvidia-*': No such file or directory
dirname: missing operand
Try 'dirname --help' for more information.
  CC [M]  /home/lexming/rpmbuild/BUILD/gdrcopy-2.2/src/gdrdrv/nv-p2p-dummy.o
/home/alex/rpmbuild/BUILD/gdrcopy-2.2/src/gdrdrv/nv-p2p-dummy.c:48:20: fatal error: nv-p2p.h: No such file or directory
 #include "nv-p2p.h"
                    ^
compilation terminated.
```

I agree with you that the `%install` step should not require `NVIDIA_SRC_DIR` as `gdrdrv.ko` is already built. I'm not sure why `driver` was added to `drv_install` in the Makefile, but maybe is just to rebuild the ko in the src directory. Nonetheles, fixing the Makefile is valid alternative.

pakmarkthub: Thank you for providing the additional information. It helps us uncover a dependency declaration problem inside `Makefile`. I have fixed it and also added `NVIDIA_SRC_DIR` to the `make drv_install` line.

FYI: We intend for users to use `build-rpm-packages.sh`. As such, we don't test `gdrcopy.spec` as a standalone script. Can you tell us why you cannot use `build-rpm-packages.sh` and want to call `gdrcopy.spec` directly?

lexming: Using the spec file integrates better with our build system. It is easier for us to directly set the `--define` options (_eg_ we have non-standard paths and we use custom `_release` strings). Our system already has a `rpmbuild` environment as well, so we prefer to build in it, instead of using a tmp dir.
Thanks for making those fixes, they worked well on our side.

drossetti (changes requested):
I believe you want to document this new feature in the README file

pakmarkthub:
> 
> 
> I believe you want to document this new feature in the README file

Thank you @drossetti. Added a few sentences about this feature to README.md.

README.md
```diff
@@ -85,6 +85,11 @@ $ sudo rpm -Uvh gdrcopy-<version>.<platform>.rpm
 $ sudo rpm -Uvh gdrcopy-devel-<version>.<platform>.rpm
 ```
 
+Non-DKMS kmod package can be generated by adding `-m` option to `build-rpm-packages.sh`.
```
drossetti, line 88: Can you rephrase here? Let's use the convention in https://forums.fedoraforum.org/showthread.php?249991-kmod-akmod-dkms, i.e. kmod and dkms are distinct ways to deliver kernel modules.

Can you also clarify that build-rpm-packages.sh builds dkms by default?

README.md
```diff
@@ -85,6 +85,11 @@ $ sudo rpm -Uvh gdrcopy-<version>.<platform>.rpm
...
+Unlike DKMS kmod, this package is specific to the NVIDIA driver version and the
```
drossetti, line 89: "unlike the dkms package, a kmod package contains a prebuilt kernel module which is specific ..."

README.md
drossetti: also, for line 77, can you specify that it is required on RPM based distro when building dkms ?
and consider moving that EPEL extra installation step to the requirements section.

drossetti (changes requested):
a few small nits

pakmarkthub:
@drossetti I have fixed the text and did some renaming. Please re-review.

drossetti (approved):

PR #195, opened 2021-04-23
Split gdrcopy deb package into libgdrapi and gdrcopy-tests packages

pakmarkthub:
Problem: see #187.

This PR:
- splits gdrcopy deb package into libgdrapi (library only) and gdrcopy-tests (test applications only) packages.
- creates gdrcopy meta-package.
- adds options to disable building gdrdrv-dkms and/or gdrcopy-tests packages. Hence, users can choose to generate only the packages they want. NVIDIA driver and/or CUDA may not be required depending on the selection.
- builds on top of PR #192. Merge that PR before this PR would be easier.

Presubmit testing:
- Installed and replaced gdrcopy-2.1-1 on gc11.


packages/debian-lib/changelog
```diff
@@ -14,7 +14,7 @@ gdrcopy (@FULL_VERSION@) stable; urgency=low
 
  -- Pak Markthub <pmarkthub@nvidia.com>  Mon, 01 Feb 2021 11:59:59 -0700
 
-gdrcopy (2.1) stable; urgency=low
```
drossetti, line 17: Does this work? Can you attribute the changelog to the gdrcopy metapackage instead?
just a question, not a blocker

pakmarkthub: It works. But I decide to also include changelog and readme for the meta package. On the other hand, changelog for libgdrapi and gdrcopy-tests packages has been set to initial version -- those packages were parts of gdrcopy package but this is the first time they exist on their own.

Can I ask you to re-review?

packages/gdrcopy.cfg
```diff
@@ -0,0 +1,17 @@
...
+### Uncomment to edit them.
+# Source: <source package name; defaults to package name>
+Section: misc
+Priority: optional
+Homepage: https://github.com/NVIDIA/gdrcopy
+Standards-Version: @FULL_VERSION@
+
+Package: gdrcopy
+Version: @FULL_VERSION@
+Maintainer: Davide Rossetti <drossetti@nvidia.com>
```
drossetti, line 11: @pakmarkthub you should be mentioned here as Maintainer

drossetti (changes requested):
just a small last nit

packages/gdrcopy.cfg
pakmarkthub: Quite reluctant to do but I have changed it.

By the way, is there a practice to put the team's name there instead. That would make it more correct.

drossetti: can we have 2 maintainers? either 2 lines or single line with 2 names separated somehow

drossetti (changes requested):
2 maintainers should be ok

pakmarkthub:
Added two maintainers for deb packages.

I don't see that concept in RPM and the changelog section accepts only one name.

drossetti (approved):

Issue #215, opened 2021-10-28
/etc/init.d/gdrcopy fails to start on SLES


bcfriesen:
Greetings,

on a SLES 15 SP1 system with gdrcopy installed via RPM, one observes the following during boot:

```console
Oct 15 19:34:26 localhost systemd[1]: Starting LSB: Startup/shutdown script for GDRcopy kernel-mode driver...
Oct 15 19:34:26 localhost gdrcopy[6233]: /etc/init.d/gdrcopy: line 17: /etc/rc.d/init.d/functions: No such file or directory
Oct 15 19:34:26 localhost gdrcopy[6233]: Checking required modules: /etc/init.d/gdrcopy: line 95: success: command not found
Oct 15 19:34:26 localhost gdrcopy[6233]: /etc/init.d/gdrcopy: line 95: failure: command not found
Oct 15 19:34:26 localhost systemd[1]: gdrcopy.service: Control process exited, code=exited status=127
Oct 15 19:34:26 localhost systemd[1]: Failed to start LSB: Startup/shutdown script for GDRcopy kernel-mode driver.
Oct 15 19:34:26 localhost systemd[1]: gdrcopy.service: Unit entered failed state.
Oct 15 19:34:26 localhost systemd[1]: gdrcopy.service: Failed with result 'exit-code'.
```

which leads to this entry in `systemctl --failed`:

```
localhost:~ # systemctl --failed
  UNIT                         LOAD   ACTIVE SUB    DESCRIPTION
● gdrcopy.service              loaded failed failed LSB: Startup/shutdown script for GDRcopy kernel-mode driver
```

The problem is that the `gdrcopy` service is trying to read `/etc/rc.d/init.d/functions`, which I believe is part of RHEL but not SLES.

This error appears to be functionally harmless, and can be worked around (one can simply delete the `/etc/init.d/gdrcopy` file), but I wonder if it might be possible to adjust the gdrcopy init.d configuration to work on SLES, or at least to avoid this constantly failed process.

Thanks very much.

drossetti:
@bcfriesen has this been fixed?

FlorianDeconinck:
Hello,

Looking to also build GDR_COPY on SLES (12 or 15). @bcfriesen did you get the chance to test the fix?

Still no official support for SLES @drossetti ?

Thank you.

cponder:
It's been almost 2 years since the last official release.
Could you tag a release-candidate with SLES support?

PR #216, opened 2021-11-09
Fixed errors in init.d/gdrcopy when running on SLES

pakmarkthub:
Problem:
- See issue #215.
- `/etc/rc.d/init.d/functions` is not available on SLES.

This PR:
- adds check if `init.d/functions` is available.
- if it isn't, we define our own `success` and `failure` functions.

Presubmit testing:
- on SLES15.2.


drossetti (approved):
LGMT

PR #223, opened 2022-02-11
Fix minor readability issue in gdr_open

al42and:
Helpfully hinted by Clang: & has lower precedence than ==; == will be evaluated first.

The result is the same: `(ps_tmp) & (0x1 == 1)` is `ps_tmp & 1` is `(ps_tmp & 0x1) == 1`, but the current form looks dangerously misleading.


pakmarkthub (approved):
I agree that we should add this protection, just in case there is a compiler that interprets it in a different order.

PR #228, opened 2022-03-14
remove build time dependency on CUDA runtime header

drossetti:
cuda_runtime_api.h is included but not really used by test apps, so let us get rid of that inclusion


drossetti:
@pakmarkthub can you take a look please?

pakmarkthub:
LGTM. Merged.

PR #229, opened 2022-04-11
Makefile: remove `ldconfig -n`.

haampie:
`ldconfig` can't be assumed to be available.

```
ldconfig -n /tmp/spack-stage/spack-stage-gdrcopy-2.3-lxw353euu26enjkcctzbux6n6cj3wjgt/spack-src/src
make[1]: ldconfig: Command not found
make[1]: *** [Makefile:60: libgdrapi.so.2.3] Error 127
```

```
cat /etc/os-release 
NAME="SLES"
VERSION="15-SP2"
VERSION_ID="15.2"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP2"
```

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -57,7 +57,6 @@ lib: $(LIB)
 $(LIBOBJS): CFLAGS+=-fPIC
 $(LIB): $(LIBOBJS)
 	$(CC) -shared -Wl,-soname,$(LIB_SONAME) -o $@ $^
-	ldconfig -n $(PWD)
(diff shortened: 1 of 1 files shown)


e4t:
You cannot remove the call to `ldconfig` here or it will come back to haunt you later when linking against `libgdrapi` when building the tests. 
The reason why it's not found is that it is located under `/sbin`. Have a look at this [patch](https://github.com/NVIDIA/gdrcopy/pull/270/commits/3f4deae4054c32d535acd5f2981851474e1144b8).

PR #248, opened 2023-03-06
[Fix issue 244] fix bug in gdr_generate_mr_handle

### LINKED ISSUE #244: gdrdrv returns the same handle for two pinned buffers
Under the following conditions, gdrdrv returns the same handle for two different pinned buffers:
1. The size of the first buffer is smaller than the GPU page.
2. The start of the second buffer is within the same page of the first buffer.

The handle for the second buffer supersedes that of the first buffer. One scenario that this behavior becomes problematic is:
- The first buffer is page-aligned.
- The second buffer is not page-aligned.

This will run correctly.
```
gdr_pin_buffer(g, first_buffer_ptr, first_buffer_size, 0, 0, &first_buffer_handle);
gdr_map(g, first_buffer_handle, &first_buffer_map_ptr, first_buffer_size);
gdr_pin_buffer(g, second_buffer_ptr, second_buffer_size, 0, 0, &second_buffer_handle);
```

This will fail because the offset of the second buffer is non-zero. But the failure happens when trying to map the first buffer!
```
gdr_pin_buffer(g, first_buffer_ptr, first_buffer_size, 0, 0, &first_buffer_handle);
gdr_pin_buffer(g, second_buffer_ptr, second_buffer_size, 0, 0, &second_buffer_handle);
gdr_map(g, first_buffer_handle, &first_buffer_map_ptr, first_buffer_size);
```

Code in #242 can be used to reproduce this bug.

pakmarkthub:
Issue:
- fixes #244.

This PR:
- introduces basic_small_buffers_mapping unit test in sanity.
- fixes bug in gdr_generate_mr_handle.
- bonus: fixes gdrdrv/Makefile to detect changes in gdrdrv.c.

Presubmit testing:
- locally on gc11 (x86).

### DIFF
--- a/src/gdrdrv/gdrdrv.c
+++ b/src/gdrdrv/gdrdrv.c
@@ -222,11 +222,11 @@ MODULE_PARM_DESC(info_enabled, "enable info tracing");
 #define GPU_PAGE_MASK    (~GPU_PAGE_OFFSET)
 
 #ifndef MAX
-#define MAX(a,b) ((a) > (b) ? a : b)
+#define MAX(a,b) ((a) > (b) ? (a) : (b))
 #endif
 
 #ifndef MIN
-#define MIN(a,b) ((a) < (b) ? a : b)
+#define MIN(a,b) ((a) < (b) ? (a) : (b))
 #endif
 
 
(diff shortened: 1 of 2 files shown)


src/gdrdrv/Makefile
```diff
@@ -45,7 +45,7 @@ REL_MINOR  := $(word 3,$(REL))
 
 all: gdrdrv.ko
 
-gdrdrv.ko:
+gdrdrv.ko: gdrdrv.c
```
drossetti, line 48: can we remove this one? it will be addressed in a separate PR

drossetti (changes requested):
one small nit

src/gdrdrv/Makefile
pakmarkthub: Ok. Removed.

drossetti (approved):
nice catch!

PR #261, opened 2023-05-19
Allow tests to build against libcuda.so stub

lukeyeager:
With this change, you can run build-deb-packages.sh with only the following packages installed (so, it's easily containerized):
* cuda-cudart-dev-12-1
* cuda-driver-dev-12-1
* cuda-nvcc-12-1

Before this change, you need to mount the libcuda from the host driver into the container (e.g. using NVIDIA_DRIVER_CAPABILITIES).

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -9,7 +9,7 @@ CHECK_LIBS := $(shell pkg-config --libs check)
 GDRAPI_INC := ../include
 GDRAPI_SRC := ../src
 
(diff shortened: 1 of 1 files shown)


drossetti:
@lukeyeager this is long due, thank you 
@pakmarkthub could you please review PR?

lukeyeager:
Ping

drossetti (approved):
looks good

PR #263, opened 2023-06-05
Fix vm_flags access violation in linux >= 6.3

pakmarkthub:
Issue:
- `vma->vm_flags` is marked as `const` in linux >= 6.3. `vm_flags_set` is introduced for modifying `vm_flags`.
- See https://github.com/torvalds/linux/commit/bc292ab00f6c7a661a8a605c714e8a148f629ef6.

This PR:
- declares `vm_flags_set` when compiling with linux < 6.3.
- always uses `vm_flags_set` to modify `vm_flags`.

Presubmit Testing:
- On an x86-64 machine with linux kernel 5.13.0-30-generic and 6.3.5.

### DIFF
--- a/src/gdrdrv/gdrdrv.c
+++ b/src/gdrdrv/gdrdrv.c
@@ -86,6 +86,19 @@ void address_space_init_once(struct address_space *mapping)
 }
 #endif
 
+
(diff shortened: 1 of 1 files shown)


drossetti:
hopefully the upstream patch will not be back ported to distro kernels

drossetti (approved):

PR #264, opened 2023-06-06
Add ping-pong latency benchmark application

pakmarkthub:
This PR:
- introduces `tests/pplat`.
- modifies the installation and packaging scripts to install pplat as `gdrcopy_pplat`.

Explanation:
- pplat measures the round-trip latency between CPU and GPU. For each iteration, CPU writes a number to GPU via `gdr_copy_to_mapping`. GPU has a CUDA kernel readily polling for this number. When the kernel observes the number, it writes this number + 1 back to the cuMemHostAlloc'd buffer. CPU polls on this cuMemHostAlloc'd buffer. The iteration is finished when CPU observes that new number from GPU.

Example output:
```
$ ./pplat -t 100000
...
selecting device 0
device ptr: 0x7face2600000
gpu alloc fn: cuMemAlloc
map_d_ptr: 0x7fad15b4f000
info.va: 7face2600000
info.mapped_size: 4
info.page_size: 65536
info.mapped: 1
info.wc_mapping: 1
page offset: 0
user-space pointer: 0x7fad15b4f000
CPU does gdr_copy_to_mapping and GPU writes back to cuMemHostAlloc'd buffer.
Running 100000 iterations with data size 4 bytes.
Round-trip latency per iteration is 1.10632 us
unmapping buffer
unpinning buffer
closing gdrdrv
```


tests/pplat.cu
```diff
@@ -0,0 +1,256 @@
...
+        cout << "page offset: " << off << endl;
+
+        d_buf = (uint32_t *)((uintptr_t)map_d_ptr + off);
+        cout << "user-space pointer: " << d_buf << endl;
+
+        cout << "CPU does gdr_copy_to_mapping and GPU writes back via cuMemHostAlloc'd buffer." << endl;
+        cout << "Running " << num_iters << " iterations with data size " << sizeof(*d_buf) << " bytes." << endl;
+
+        pp_kernel<<< 1, 1 >>>((uint32_t *)d_buf_cuptr, (uint32_t *)h_buf_cuptr, num_iters);
+
```
drossetti, line 210: I would check for early async errors with ASSERTDRV(cuStreamQuery(0));

tests/pplat.cu
```diff
@@ -0,0 +1,256 @@
...
+        uint32_t i = 1;
+        while (READ_ONCE(*h_buf) != i) ;
+        MB();
+
+        clock_gettime(MYCLOCK, &beg);
+        while (i < num_iters) {
```
drossetti, line 216: I'd add some kind of timeout check, possibly every X iterations to limit the overhead.
On timeout, I'd call ASSERTDRV(cuStreamQuery(0)); again to check and report possible for async errors.

tests/pplat.cu
```diff
@@ -0,0 +1,256 @@
...
+
+        cout << "unpinning buffer" << endl;
+        ASSERT_EQ(gdr_unpin_buffer(g, mh), 0);
+    } END_CHECK;
+
+    cout << "closing gdrdrv" << endl;
+    ASSERT_EQ(gdr_close(g), 0);
+
+    ASSERTDRV(gfree_fn(&mhandle));
+
```
drossetti, line 245: we may want to free h_buf

tests/pplat.cu
```diff
@@ -0,0 +1,256 @@
...
+        cout << "user-space pointer: " << d_buf << endl;
+
+        cout << "CPU does gdr_copy_to_mapping and GPU writes back via cuMemHostAlloc'd buffer." << endl;
+        cout << "Running " << num_iters << " iterations with data size " << sizeof(*d_buf) << " bytes." << endl;
+
+        pp_kernel<<< 1, 1 >>>((uint32_t *)d_buf_cuptr, (uint32_t *)h_buf_cuptr, num_iters);
+
+        uint32_t i = 1;
+        while (READ_ONCE(*h_buf) != i) ;
+        MB();
```
drossetti, line 213: MB on Arm64 seems to be just a compiler barrier. Don't we need a stronger form?

drossetti (changes requested):

tests/pplat.cu
pakmarkthub: Nice catch! Yes, I updated the memory barrier instruction for ARM.

tests/pplat.cu
pakmarkthub: Indeed. Added.

tests/pplat.cu
pakmarkthub: Added.

tests/pplat.cu
```diff
@@ -0,0 +1,256 @@
...
+            MB();
+
+            ++i;
+
+            while (READ_ONCE(*h_buf) != i) ;
+            MB();
+        }
+        clock_gettime(MYCLOCK, &end);
+
+        ASSERTRT(cudaDeviceSynchronize());
```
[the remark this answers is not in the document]
pakmarkthub: Changed to `cuStreamSynchronize`.

tests/pplat.cu
pakmarkthub: Added.

drossetti (changes requested):

drossetti:
README needs to be updated too

tests/pplat.cu
drossetti: > CUDA_ERROR_NOT_READY = 600
> This indicates that asynchronous operations issued previously have not completed yet. This result is not actually an error, but must be indicated differently than [CUDA_SUCCESS](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ggc6c391505e117393cc2558fff6bfc2e9a0eed720f8a87cd1c5fd1c453bc7a03d) (which indicates completion). Calls that may return this value include [cuEventQuery()](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EVENT.html#group__CUDA__EVENT_1g6f0704d755066b0ee705749ae911deef) and [cuStreamQuery()](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__STREAM.html#group__CUDA__STREAM_1g1b0d24bbe97fa68e4bc511fb6adfeb0b).

So we exactly want to verify that the kernel has been launched, so we should assert that cuStreamQuery(0) returns CUDA_ERROR_NOT_READY at that point.

tests/pplat.cu
```diff
@@ -53,21 +53,46 @@ __global__ void pp_kernel(uint32_t *d_buf, uint32_t *h_buf, uint32_t num_iters)
...
+static inline void check_timeout(struct timespec start, double timeout_us)
+{
+    struct timespec now;
+    double time_used_us;
+    if (timeout_us > 0) {
+        clock_gettime(MYCLOCK, &now);
+        time_used_us = time_diff(start, now);
+        if (time_used_us > timeout_us) {
+            cerr << "ERROR: TIMEOUT!!!" << endl;
+            ASSERTDRV(cuStreamQuery(0));
```
drossetti, line 90: - CUDA_ERROR_NOT_READY would mean that the kernel is  still running, so that would be a timeout caused by a hanging kernel
- CUDA_SUCCESS would mean that the kernel has exited, which is still unexpected, so it is not a good exit value either
- any other error would be unexpected.
I would just print and decode any code returned here

tests/pplat.cu
```diff
@@ -209,25 +240,32 @@ int main(int argc, char *argv[])
         pp_kernel<<< 1, 1 >>>((uint32_t *)d_buf_cuptr, (uint32_t *)h_buf_cuptr, num_iters);
 
         uint32_t i = 1;
-        while (READ_ONCE(*h_buf) != i) ;
+        // Wait for pp_kernel to be ready before starting the time measurement.
+        clock_gettime(MYCLOCK, &beg);
+        while (READ_ONCE(*h_buf) != i) {
+            check_timeout(beg, timeout_us);
```
drossetti, line 246: I would check for timeout every 10^3 iterations or so

drossetti (changes requested):

tests/pplat.cu
drossetti: so that you can turn it on by default with a large value, say 1 second, without affecting the latency measurement

tests/pplat.cu
pakmarkthub: Added.

config_arch
```diff
@@ -19,6 +19,7 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
+scriptdir="$(cd "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")" && pwd)"
```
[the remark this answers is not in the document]
pakmarkthub: `config_arch` is a script. So, it is the directory where this script is located. Basically, this change allows us to call `config_arch` from anywhere.

tests/Makefile
```diff
@@ -44,6 +46,12 @@ copylat: copylat.o common.o
 apiperf: apiperf.o common.o
 	$(LINK.cc)  -o $@ $^ $(LIBS) -lrt
 
+pplat: pplat.o common.o
+	$(LINK.cc)  -o $@ $^ $(LIBS) -lrt -lcudart
```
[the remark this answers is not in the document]
pakmarkthub: Thank you for the suggestion. Changed to libcuda_static.

tests/pplat.cu
pakmarkthub: Changed the code to report status returned from `cuStreamQuery(0)`.

tests/pplat.cu
pakmarkthub: Implemented.

pakmarkthub:
> README needs to be updated too

Updated.

tests/pplat.cu
```diff
@@ -0,0 +1,312 @@
...
+        d_buf = (uint32_t *)((uintptr_t)map_d_ptr + off);
+        cout << "user-space pointer: " << d_buf << endl;
+
+        cout << "CPU does gdr_copy_to_mapping and GPU writes back via cuMemHostAlloc'd buffer." << endl;
+        cout << "Running " << num_iters << " iterations with data size " << sizeof(*d_buf) << " bytes." << endl;
+
+        pp_kernel<<< 1, 1 >>>((uint32_t *)d_buf_cuptr, (uint32_t *)h_buf_cuptr, num_iters);
+
+        // Catching any potential errors. CUDA_ERROR_NOT_READY means pp_kernel
+        // is running. We expect to see this status than CUDA_SUCCESS because
```
drossetti, line 254: "than" -> "instead of"

tests/common.hpp
```diff
@@ -29,16 +29,39 @@
...
 #elif defined(GDRAPI_POWER)
 #define MB() asm volatile("sync":::"memory")
+#define SB() MB()
+#define LB() MB()
+#elif defined(GDRAPI_ARM64)
+#define MB() asm volatile("dmb sy":::"memory")
+#define SB() asm volatile("dmb st":::"memory")
+#define LB() MB()
 #else
 #define MB() asm volatile("":::"memory")
```
drossetti, line 62: shall we `#error` here on an unsupported platform?

config_arch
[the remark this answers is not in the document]
drossetti: I would still call it sourcedir or similarly, since you are using it below to pass the include paths to gcc

drossetti (changes requested):

tests/common.hpp
pakmarkthub: Good idea. Added.

config_arch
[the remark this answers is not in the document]
pakmarkthub: A nitpick. I don't like the name `sourcedir` because it misleads me to think about the `src` folder in GDRCopy. I propose that we use `topdir` instead.

drossetti (approved):

drossetti:
this PR needs to be rebased

pakmarkthub:
Thank you. Rebased and merged.

Issue #262, opened 2023-05-30
Ubuntu 22.04 - gdrcopy 2.3.1 dependency problem


YoonGi-AWS:
When I try to install gdrcopy 2.3.1 into Ubuntu 22.04(5.19.0-1025-aws), I faced dependency problem.

```
root@ip-172-31-18-103:/home/ubuntu/gdrcopy-2.3.1/packages# dpkg -i gdrdrv-dkms_*_amd64.*.deb
Selecting previously unselected package gdrdrv-dkms:amd64.
(Reading database ... 100387 files and directories currently installed.)
Preparing to unpack gdrdrv-dkms_2.3.1_amd64.Ubuntu22_04.deb ...
Unpacking gdrdrv-dkms:amd64 (2.3.1) ...
Setting up gdrdrv-dkms:amd64 (2.3.1) ...
Loading new gdrdrv-2.3.1 DKMS files...
Building for 5.19.0-1025-aws
Building for architecture x86_64
Building initial module for 5.19.0-1025-aws
EFI variables are not supported on this system
/sys/firmware/efi/efivars not found, aborting.
Done.

gdrdrv.ko:
Running module version sanity check.
 - Original module
   - No original module exists within this kernel
 - Installation
   - Installing to /lib/modules/5.19.0-1025-aws/updates/dkms/

depmod...


root@ip-172-31-18-103:/home/ubuntu/gdrcopy-2.3.1/packages# dpkg -i libgdrapi_*_amd64.*.deb
Selecting previously unselected package libgdrapi:amd64.
(Reading database ... 100396 files and directories currently installed.)
Preparing to unpack libgdrapi_2.3.1_amd64.Ubuntu22_04.deb ...
Unpacking libgdrapi:amd64 (2.3.1) ...
Setting up libgdrapi:amd64 (2.3.1) ...
Processing triggers for libc-bin (2.35-0ubuntu3.1) ...


root@ip-172-31-18-103:/home/ubuntu/gdrcopy-2.3.1/packages# dpkg -i gdrcopy-tests_*_amd64.*.deb
Selecting previously unselected package gdrcopy-tests:amd64.
(Reading database ... 100404 files and directories currently installed.)
Preparing to unpack gdrcopy-tests_2.3.1_amd64.Ubuntu22_04.deb ...
Unpacking gdrcopy-tests:amd64 (2.3.1) ...
Setting up gdrcopy-tests:amd64 (2.3.1) ...


[...]

PR #265, opened 2023-06-20
Fixed the dependency declaration in debian-meta/control

pakmarkthub:
Problem:
- See #262.

This PR:
- changes the Depends field of debian-meta to `@VERSION@`.

Presubmit Testing:
- on Ubuntu 20.04.

### DIFF
--- a/packages/debian-meta/control
+++ b/packages/debian-meta/control
@@ -10,7 +10,7 @@ Homepage: https://github.com/NVIDIA/gdrcopy
 Package: gdrcopy
 Architecture: any
 Multi-Arch: same
-Depends: gdrdrv-dkms (= @FULL_VERSION@), libgdrapi (= @FULL_VERSION@), gdrcopy-tests (= @FULL_VERSION@)
+Depends: gdrdrv-dkms (= @VERSION@), libgdrapi (= @VERSION@), gdrcopy-tests (= @VERSION@)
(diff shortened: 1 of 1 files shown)


drossetti:
Have you had a chance to test on Ubuntu 22.04, as reported on #262?
Any idea why QA team has not seen this problem?

drossetti:
It seems that you want to submit to R2.3.1, which was supposed to be a quick release.
I'd rather submit to main/master and cherry-pick it where necessary.

pakmarkthub:
This issue is specific to R2.3.1. Except this quick release, our version should be 2.x, not 2.x.y. And because this is a quick release, we didn't complete the entire test plan.

I could reproduce this issue on Ubuntu 20.04. And this PR fixes this issue there. Regardless, I will try it on Ubuntu 22.04 also.

drossetti:
> This issue is specific to R2.3.1. Except this quick release, our version should be 2.x, not 2.x.y. And because this is a quick release, we didn't complete the entire test plan.
> 
> I could reproduce this issue on Ubuntu 20.04. And this PR fixes this issue there. Regardless, I will try it on Ubuntu 22.04 also.

Ah I see, the bug has to do with the subversion. So do you plan to tag a 2.3.2?

Regardless I would have this fix on the master branch too.

drossetti (approved):

PR #270, opened 2023-07-21
Add Support for SLE and openSUSE Leap

e4t:
This PR:
-  Adds support for openSUSE/SUSE distributions:
   - Fixes search path to NVIDA kernel driver bits.
   - Correctly sets release_version for openSUSE/SUSE in RPM spec file,
   - Creates `modprobe` configuration to set up device,
   - Adds instructions to README how to build for openSUSE/SUSE.
- Makes sure, architecture is correctly appended as as last element in rpm package names.
- Makes sure, `ldconfig` is found even when building as user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,8 @@ The device driver requires GPU display driver >= 418.40 on ppc64le and >= 331.14
 require CUDA >= 6.0. Additionally, the _sanity_ test requires check >= 0.9.8 and
 subunit.
 
-DKMS is a prerequisite for installing GDRCopy kernel module package. On RHEL,
+DKMS is a prerequisite for installing GDRCopy kernel module package. On RHEL
+or SLE,
 however, users have an option to build kmod and install it instead of the DKMS
 package. See [Build and installation](#build-and-installation) section for more details.
 
@@ -62,6 +63,10 @@ $ sudo yum install dkms check check-devel subunit subunit-devel
 
 # On Debian
 $ sudo apt install check libsubunit0 libsubunit-dev
+
+# On SLE / Leap
+# On SLE dkms can be installed from PackageHub.
+$ sudo zypper install dkms check-devel rpmbuild
 ```
 
 CUDA and GPU display driver must be installed before building and/or installing GDRCopy.
@@ -72,9 +77,11 @@ of the driver (or CUDA) installation with  *runfile*. If you install the driver
 via package management, we suggest
 - On RHEL, `sudo dnf module install nvidia-driver:latest-dkms`.
 - On Debian, `sudo apt install nvidia-dkms-<your-nvidia-driver-version>`.
+- On SLE, `sudo zypper install nvidia-gfx<your-nvidia-driver-version>-kmp`.
 
 The supported architectures are Linux x86_64, ppc64le, and arm64. The supported
-platforms are RHEL7, RHEL8, Ubuntu16_04, Ubuntu18_04, and Ubuntu20_04.
+platforms are RHEL7, RHEL8, Ubuntu16_04, Ubuntu18_04, Ubuntu20_04,
+SLE-15 (any SP) and Leap 15.x.
 
 Root privileges are necessary to load/install the kernel-mode device
 driver.
@@ -87,8 +94,13 @@ We provide three ways for building and installing GDRCopy.
 ### rpm package
 
 ```shell
+# For RHEL:
 $ sudo yum groupinstall 'Development Tools'
 $ sudo yum install dkms rpm-build make check check-devel subunit subunit-devel
+
+# For SLE:
+$ sudo zypper in dkms rpmbuild check-devel
+
 $ cd packages
 $ CUDA=<cuda-install-top-dir> ./build-rpm-packages.sh
 $ sudo rpm -Uvh gdrcopy-kmod-<version>dkms.noarch.<platform>.rpm
--- a/packages/build-rpm-packages.sh
+++ b/packages/build-rpm-packages.sh
@@ -101,7 +101,7 @@ FULL_VERSION="${VERSION}"
 
 if [[ ${generate_kmod} == 1 ]]; then
     if [ -z "${NVIDIA_SRC_DIR}" ]; then
-        NVIDIA_SRC_DIR=$(find /usr/src/nvidia-* -name "nv-p2p.h" -print -quit)
+	NVIDIA_SRC_DIR=$(find /usr/src/kernel-modules/nvidia-* /usr/src/nvidia-* -name "nv-p2p.h" -print -quit 2>/dev/null)
         if [ ${#NVIDIA_SRC_DIR} -gt 0 ]; then
             NVIDIA_SRC_DIR=$(dirname ${NVIDIA_SRC_DIR})
         fi
--- a/packages/build-rpm-packages.sh
+++ b/packages/build-rpm-packages.sh
@@ -160,6 +160,8 @@ if [ -f "/etc/redhat-release" ]; then
     release_version=".el$(cat /etc/redhat-release | grep -o -E '[0-9]+' | head -1)"
 elif [ -f "/etc/centos-release" ]; then
     release_version=".el$(cat /etc/centos-release | grep -o -E '[0-9]+' | head -1)"
+elif [ -f "/etc/os-release" ]; then
+    release_version=$(source /etc/os-release && echo ".$ID-$VERSION_ID")
 else
     release_version="unknown_distro"
 fi
--- a/packages/build-rpm-packages.sh
+++ b/packages/build-rpm-packages.sh
@@ -168,12 +168,13 @@ fi
 echo $srpm $rpms
 ex cd ${CWD}
 for item in `ls $tmpdir/topdir/SRPMS/*.rpm $tmpdir/topdir/RPMS/*/*.rpm`; do
-    item_name=`basename $item`
-    item_name=`echo $item_name | sed -e "s/\.rpm//g"`
+    item_name=`basename $item .rpm`
+    arch=$(sed -ne 's/.*\(\.[^\.]\+\)$/\1/p' <<< $item_name)
+    item_name=`basename $item_name $arch`
     if [ "$item_name" == "gdrcopy-${FULL_VERSION}-${RPM_VERSION}.`uname -m`" ]; then
-        item_name="${item_name}${release_version}+cuda${CUDA_MAJOR}.${CUDA_MINOR}.rpm"
+        item_name="${item_name}${release_version}+cuda${CUDA_MAJOR}.${CUDA_MINOR}.${arch}.rpm"
     else
-        item_name="${item_name}${release_version}.rpm"
+        item_name="${item_name}${release_version}${arch}.rpm"
     fi
     ex cp $item ./${item_name}
 done
--- a/packages/gdrcopy.spec
+++ b/packages/gdrcopy.spec
@@ -150,9 +150,20 @@ cp -a $RPM_BUILD_DIR/%{name}-%{version}/src/gdrdrv/Makefile $RPM_BUILD_ROOT%{usr
 cp -a $RPM_BUILD_DIR/%{name}-%{version}/src/gdrdrv/nv-p2p-dummy.c $RPM_BUILD_ROOT%{usr_src_dir}/gdrdrv-%{version}/
 cp -a $RPM_BUILD_DIR/%{name}-%{version}/dkms.conf $RPM_BUILD_ROOT%{usr_src_dir}/gdrdrv-%{version}
 
+%if 0%{!?suse_version:1}
 # Install gdrdrv service script
 install -d $RPM_BUILD_ROOT/etc/init.d
 install -m 0755 $RPM_BUILD_DIR/%{name}-%{version}/init.d/gdrcopy $RPM_BUILD_ROOT/etc/init.d
+%else
+mkdir -p $RPM_BUILD_ROOT/etc/modprobe.d
+cat <<"EOF" > $RPM_BUILD_ROOT/etc/modprobe.d/50-gdrdrv.conf
+#options gdrdrv dbg_enabled=1
+#options info_enabled=1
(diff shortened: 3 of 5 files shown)


packages/gdrcopy.spec
```diff
@@ -150,9 +150,19 @@ cp -a $RPM_BUILD_DIR/%{name}-%{version}/src/gdrdrv/Makefile $RPM_BUILD_ROOT%{usr
...
+%if 0%{!?suse_version:1}
 # Install gdrdrv service script
 install -d $RPM_BUILD_ROOT/etc/init.d
 install -m 0755 $RPM_BUILD_DIR/%{name}-%{version}/init.d/gdrcopy $RPM_BUILD_ROOT/etc/init.d
+%else
+mkdir -p $RPM_BUILD_ROOT/etc/modprobe.d
+cat <<"EOF" > $RPM_BUILD_ROOT/etc/modprobe.d/50-gdrdrv.conf
+#options gdrdrv dbg_enabled=1
+#options info_enabled=1
+install gdrdrv PATH=$PATH:/bin:/usr/bin; /sbin/modprobe -i gdrdrv && rm -f /dev/gdrdrv && mknod -m 666 /dev/gdrdrv c $(gawk '/gdrdrv/{printf"%%s",$1}' /proc/devices) 0 && chgrp video /dev/gdrdrv
```
drossetti, line 162: just wondering if you might need `--ignore-install` here.

e4t: Shouldn't  `-i` and `--ignore-install` should essentially be the same in this context? Since the man page explicitly uses it in its example I will replace it. $CMDLINE_OPTS should definitely be added.

README.md
```diff
@@ -72,9 +77,11 @@ of the driver (or CUDA) installation with  *runfile*. If you install the driver
 via package management, we suggest
 - On RHEL, `sudo dnf module install nvidia-driver:latest-dkms`.
 - On Debian, `sudo apt install nvidia-dkms-<your-nvidia-driver-version>`.
+- On SLE, `sudo zypper install nvidia-gfx<your-nvidia-driver-version>-kmp`.
 
 The supported architectures are Linux x86_64, ppc64le, and arm64. The supported
-platforms are RHEL7, RHEL8, Ubuntu16_04, Ubuntu18_04, and Ubuntu20_04.
+platforms are RHEL7, RHEL8, Ubuntu16_04, Ubuntu18_04, Ubuntu20_04,
+SLE-15 (any SP) and Leap 15.x.
```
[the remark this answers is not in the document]
e4t: For that matter, SLE 15 <N> and Leap 15.<N> are essentially the same. We will make efforts to test GDRcopy as part of our product release process.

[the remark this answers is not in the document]
drossetti: That is very much appreciated.
In any case, the tests included in this project build using the CUDA toolkit, which has its own version-dependent support matrix, see https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#os-support-policy.
So at any given time, you will only be able to build on a distro version for which a current/old CUDA toolkit has support.

packages/gdrcopy.spec
```diff
@@ -150,9 +150,19 @@ cp -a $RPM_BUILD_DIR/%{name}-%{version}/src/gdrdrv/Makefile $RPM_BUILD_ROOT%{usr
...
 # Install gdrdrv service script
 install -d $RPM_BUILD_ROOT/etc/init.d
 install -m 0755 $RPM_BUILD_DIR/%{name}-%{version}/init.d/gdrcopy $RPM_BUILD_ROOT/etc/init.d
+%else
+mkdir -p $RPM_BUILD_ROOT/etc/modprobe.d
+cat <<"EOF" > $RPM_BUILD_ROOT/etc/modprobe.d/50-gdrdrv.conf
+#options gdrdrv dbg_enabled=1
+#options info_enabled=1
+install gdrdrv PATH=$PATH:/bin:/usr/bin; /sbin/modprobe --ignore-install gdrdrv $CMDLINE_OPTS && rm -f /dev/gdrdrv && mknod -m 666 /dev/gdrdrv c $(gawk '/gdrdrv/{printf"%%s",$1}' /proc/devices) 0 && chgrp video /dev/gdrdrv
+remove gdrdrv PATH=$PATH:/bin:/usr/bin; /sbin/modprobe -ir gdrdrv && rm -f /dev/gdrdrv
```
drossetti, line 163: since you used `--ignore-install` above, you could use `--ignore-remove --remove` here, instead of the terser `-ir`

packages/gdrcopy.spec
```diff
@@ -150,9 +150,19 @@ cp -a $RPM_BUILD_DIR/%{name}-%{version}/src/gdrdrv/Makefile $RPM_BUILD_ROOT%{usr
...
+EOF
+%endif
```
drossetti, line 165: do you need to set non-default permissions on 50-gdrdrv.conf?

packages/gdrcopy.spec
drossetti: you are right, `-i` and `--ignore-install` are synonymous.

packages/build-rpm-packages.sh
```diff
@@ -101,7 +101,7 @@ FULL_VERSION="${VERSION}"
 
 if [[ ${generate_kmod} == 1 ]]; then
     if [ -z "${NVIDIA_SRC_DIR}" ]; then
-        NVIDIA_SRC_DIR=$(find /usr/src/nvidia-* -name "nv-p2p.h" -print -quit)
+        NVIDIA_SRC_DIR=$(find $(find /usr/src/ -type d -a -name nvidia-* -prune) -name "nv-p2p.h" -print -quit)
```
pakmarkthub, line 104: If `$(find /usr/src/ -type d -a -name nvidia-* -prune)` outputs blank (cannot find any nvidia source code), the outer `find` will search the current directory and its sub directories. It could have unintended result if the current directory or its sub directory contains nv-p2p.h.

packages/build-rpm-packages.sh
```diff
@@ -160,18 +160,22 @@ if [ -f "/etc/redhat-release" ]; then
     release_version=".el$(cat /etc/redhat-release | grep -o -E '[0-9]+' | head -1)"
 elif [ -f "/etc/centos-release" ]; then
     release_version=".el$(cat /etc/centos-release | grep -o -E '[0-9]+' | head -1)"
+elif [ -f "/etc/os-release" ]; then
+    . /etc/os-release
```
pakmarkthub, line 164: It is not a good idea to `source /etc/os-release` here without protection. It could lead to conflicting variables later. For example, what if we use `$ID` or `$VERSION_ID` in the future? Are there other variables that /etc/os-release might replace?

packages/gdrcopy.spec
```diff
@@ -150,9 +150,19 @@ cp -a $RPM_BUILD_DIR/%{name}-%{version}/src/gdrdrv/Makefile $RPM_BUILD_ROOT%{usr
...
+%if 0%{!?suse_version:1}
 # Install gdrdrv service script
 install -d $RPM_BUILD_ROOT/etc/init.d
 install -m 0755 $RPM_BUILD_DIR/%{name}-%{version}/init.d/gdrcopy $RPM_BUILD_ROOT/etc/init.d
+%else
+mkdir -p $RPM_BUILD_ROOT/etc/modprobe.d
+cat <<"EOF" > $RPM_BUILD_ROOT/etc/modprobe.d/50-gdrdrv.conf
+#options gdrdrv dbg_enabled=1
+#options info_enabled=1
+install gdrdrv PATH=$PATH:/bin:/usr/bin; /sbin/modprobe --ignore-install gdrdrv $CMDLINE_OPTS && rm -f /dev/gdrdrv && mknod -m 666 /dev/gdrdrv c $(gawk '/gdrdrv/{printf"%%s",$1}' /proc/devices) 0 && chgrp video /dev/gdrdrv
```
pakmarkthub, line 162: Why do you do `chgrp video /dev/gdrdrv` here? Is this the most appropriate group in SLE? Is it guaranteed to always be available?

src/gdrdrv/Makefile
```diff
@@ -18,7 +18,7 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-NVIDIA_SRC_DIR ?= $(shell find /usr/src/nvidia-* -name "nv-p2p.h"|head -1|xargs dirname || echo "NVIDIA_DRIVER_MISSING")
+NVIDIA_SRC_DIR ?= $(shell find $$(find /usr/src/ -type d -a -name "nvidia-*" -prune) -name "nv-p2p.h" -print -quit |xargs dirname || echo "NVIDIA_DRIVER_MISSING")
```
pakmarkthub, line 21: See my comment in packages/build-rpm-packages.sh.

pakmarkthub (changes requested):

packages/build-rpm-packages.sh
e4t: I will submit a patch with more safeguards.

packages/build-rpm-packages.sh
e4t: Ok, I've looked into this some more, it seems there are only two possible location of the nvidia-* directory:
/usr/src and /usr/src/kernel-modules. So the test can be much more specific.

src/gdrdrv/Makefile
e4t: See above.

packages/gdrcopy.spec
e4t: Will do.

packages/gdrcopy.spec
e4t: Not sure what you mean, permissions on /etc/modprobe.d/50-gdrdrv.conf should be 0644.

packages/build-rpm-packages.sh
```diff
@@ -101,10 +101,15 @@ FULL_VERSION="${VERSION}"
 
 if [[ ${generate_kmod} == 1 ]]; then
     if [ -z "${NVIDIA_SRC_DIR}" ]; then
-        NVIDIA_SRC_DIR=$(find /usr/src/nvidia-* -name "nv-p2p.h" -print -quit)
-        if [ ${#NVIDIA_SRC_DIR} -gt 0 ]; then
-            NVIDIA_SRC_DIR=$(dirname ${NVIDIA_SRC_DIR})
-        fi
+	for d in /usr/src/kernel-modules /usr/src; do
+	    -d $d && SEARCH_DIR=$dir && break
```
drossetti, line 105: I believe you meant `test -d`

packages/gdrcopy.spec
drossetti: For example, in the RH case, `install -m 0755 blabla`.

In the SUSE case, you use cat to create `50-gdrdrv.conf`, which will have default permissions. I was suggesting to explicitly set the permissions after creating the file.

packages/build-rpm-packages.sh
e4t: Indeed.

packages/gdrcopy.spec
pakmarkthub: On other platforms such as Ubuntu, `/dev/nvidia<N>` are set to `root:root` with 666 permission. We do the same with `/dev/gdrdrv`. I am not sure about on SUSE, but I suggest that you use the same permission and group as `/dev/nvidia<N>`. Note that users need both read and write permissions in order to access gdrdrv. In this case, 664 is the same as 660.

src/gdrdrv/Makefile
```diff
@@ -18,7 +18,10 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-NVIDIA_SRC_DIR ?= $(shell find /usr/src/nvidia-* -name "nv-p2p.h"|head -1|xargs dirname || echo "NVIDIA_DRIVER_MISSING")
+NVIDIA_SRC_DIR ?= $(shell d=/usr/src; f=''; \
```
pakmarkthub, line 21: The code is getting more complex. How about this?
```
NVIDIA_SRC_DIR ?= $(shell (find /usr/src/kernel-modules/nvidia-* /usr/src/nvidia-* -name "nv-p2p.h" -print -quit |head -1|xargs dirname || echo "NVIDIA_DRIVER_MISSING") 2>/dev/null )
```

pakmarkthub (changes requested):

src/gdrdrv/Makefile
e4t: That's what I had first, I found it ugly as `find` will most likely report an error (which gets supressed). Trying to find a way around this eventually lead to a slightly over-engineered  variant :/ 
With `-print -quit` it should be possible to drop the `head -1` though.

packages/build-rpm-packages.sh
```diff
@@ -101,10 +101,15 @@ FULL_VERSION="${VERSION}"
 
 if [[ ${generate_kmod} == 1 ]]; then
     if [ -z "${NVIDIA_SRC_DIR}" ]; then
-        NVIDIA_SRC_DIR=$(find /usr/src/nvidia-* -name "nv-p2p.h" -print -quit)
-        if [ ${#NVIDIA_SRC_DIR} -gt 0 ]; then
-            NVIDIA_SRC_DIR=$(dirname ${NVIDIA_SRC_DIR})
-        fi
+	for d in /usr/src/kernel-modules /usr/src; do
+	    [ -d $d ] && SEARCH_DIR=$dir && break
```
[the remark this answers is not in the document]
e4t: If we decide for the 'ugly variant' in the Makefile for the kernel module, we may want to do the same here - and instead suppress potential complaints by `find`.  Any variant will pick the first location found.

packages/gdrcopy.spec
e4t: On SUSE, `/dev/nvidia*`  is `root:video` and 660. I will change the /dev/gdrdrv permissions to 660 as well.

packages/gdrcopy.spec
drossetti: thanks for the change

drossetti (approved):

pakmarkthub (approved):
Thank you for your contributions. Very appreciated!

PR #271, opened 2023-07-28
Correct naming error on gdrcopy_pplat at install target

sayarsoft:
I encountered an error during installation and patched the error on naming of gdrcopy_pplat executable

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -62,7 +62,7 @@ install: exes
 	install -D -v -m u=rwx,g=rx,o=rx gdrcopy_copylat -t $(DESTBIN) && \
 	install -D -v -m u=rwx,g=rx,o=rx gdrcopy_apiperf -t $(DESTBIN) && \
 	install -D -v -m u=rwx,g=rx,o=rx gdrcopy_sanity -t $(DESTBIN) && \
-	install -D -v -m u=rwx,g=rx,o=rx pplat -T $(DESTBIN)/gdrcopy_pplat
+	install -D -v -m u=rwx,g=rx,o=rx gdrcopy_pplat -t $(DESTBIN)
 	cd $(DESTBIN) && \
 	ln -sf gdrcopy_copybw copybw && \
 	ln -sf gdrcopy_copylat copylat && \


tests/Makefile
```diff
@@ -62,11 +62,12 @@ install: exes
...
 	install -D -v -m u=rwx,g=rx,o=rx gdrcopy_sanity -t $(DESTBIN) && \
-	install -D -v -m u=rwx,g=rx,o=rx pplat -T $(DESTBIN)/gdrcopy_pplat
+	install -D -v -m u=rwx,g=rx,o=rx gdrcopy_pplat -t $(DESTBIN)
 	cd $(DESTBIN) && \
 	ln -sf gdrcopy_copybw copybw && \
 	ln -sf gdrcopy_copylat copylat && \
 	ln -sf gdrcopy_apiperf apiperf && \
-	ln -sf gdrcopy_sanity sanity
+	ln -sf gdrcopy_sanity sanity && \
+	ln -sf gdrcopy_pplat pplat
```
pakmarkthub, line 71: We want to prefix all applications with `gdrcopy_`. However, `copybw`, `copylat`, `apiperf`, and `sanity` are released before we enforce this behavior. That is why we create symlinks for users who still use those names.

`gdrcopy_pplat` was introduced after this prefix enforcement. We don't need to symlink to `pplat`. Can you remove this line?

pakmarkthub (changes requested):
Nice catch. Thank you!

tests/Makefile
sayarsoft: Thanks for the review. Sure I can do. But a second thought, this will give the option of using pplat executable both as gdrcopy_pplat or pplat as the other executables. Beside any other developer who see this Makefile probably will think pplat is missing on this link creation part like I did.
Waiting for your comment to do the necessary.

pakmarkthub: Those applications' names are common and might collide with applications from other packages. We have changed README to include `gdrcopy_` as prefix to all applications. So, the symlinks are provided as backward compatibility only. We will discuss within the team whether we should symlink `pplat` or not. But for now, please remove this symlink.

I also noticed that `pplat` is still mentioned in README. I'll create a new PR to change that to `gdrcopy_pplat`.

sayarsoft: Thanks.
I'm removing the line.

sayarsoft: Updated the commit.
Thanks for applying it.

pakmarkthub (approved):
Thank you for your contribution!
Approved.

PR #280, opened 2023-10-12
Remove check header dependency + include string

ivankochin:
If I understand correctly check lib dependency was removed several month ago, but common.hpp file still contains header inclusion. When I removed this header, I got the following errors on Ubuntu 20.04:

image

So I decided to fix two problems by one patch.


drossetti:
@ivankochin Thank you for this patch. 
FYI we had a similar fix in our pipeline which has been merged via 8f61796c9596.
So this is not needed anymore.

Issue #283, opened 2023-11-21
Driver flavor detection fails for 545 series


bsteinb:
In driver series 545, the `MODULE_LICENSE("Dual MIT/GPL")` declaration has been moved to `nv.c` while `nv-frontend.c` has been removed. Thus, the detection mechanism in `src/gdrdrv/Makefile` fails to detect the driver flavor correctly.

https://github.com/NVIDIA/open-gpu-kernel-modules/commit/b5bf85a8e3eb2516b9abca4d1becaf1172d62822#diff-cfa223e43b41102b8f71ac8d5034f4bca83f31d11a8ba1f8e4fd486e2c44043b

pakmarkthub:
The fix has been merged. Closing this issue.

bsteinb:
I thought this fix had made it into the 2.4.1 release, but apparently it has not and still persists in the latest release. I think driver version 545 was never officially released, but driver 550.54.14 was released a few days ago and gdrcopy has the same issue with detecting its flavor.

PR #284, opened 2023-11-22
Fix issue 283: NVIDIA_IS_OPENSOURCE is not set correctly with NVIDIA driver version 545

pakmarkthub:
Issue:
- See #283.
- Starting from NVIDIA driver version 545, `MODULE_LICENSE` is moved to `nv.c`.

This PR:
- includes `$(NVIDIA_SRC)/nv.c` to the search path.

Presubmit Testing:
- With proprietary flavor and opensource flavor of version 545.
- With driver version 520 to check regression.


realarnavgoel (approved):
Looks good to me.

PR #294, opened 2024-02-27
Fix building issues on OpenSUSE

JieRen98:
the [config_arch](https://github.com/NVIDIA/gdrcopy/blob/master/config_arch) script is for determining the architecture of the system. However, it generates a temp dir at `/tmp` using `mktemp` in [this line](https://github.com/NVIDIA/gdrcopy/blob/9ecd9cfa549cdc9d36785bc472d45a154f2ae7f3/config_arch#L23) and then compile and run an executable. However, a lot of Linux distributions, e.g., OpenSUSE, mount the /tmp directory with the `noexec` flag. So it is not allowed to run any executables in `/tmp`.

This PR generates the temp dir at the current dir rather than in `/tmp`.

### DIFF
--- a/config_arch
+++ b/config_arch
@@ -19,8 +19,45 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
+# Function to check directory writability and noexec flag
+check_dir() {
+    local dir=$1
+    local mount_point
+
+    # Check if directory exists and is writable
+    if [[ -d "$dir" && -w "$dir" ]]; then
+        # Find the mount point for the directory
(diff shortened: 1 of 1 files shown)


pakmarkthub:
Thank you, @JieRen98. However, this does not solve the fundamental issue. On some clusters, the current directory is immutable inside compute nodes (i.e., it is read only). Introducing a variable to Makefile to set the temp directory location might be better.

JieRen98:
> Thank you, @JieRen98. However, this does not solve the fundamental issue. On some clusters, the current directory is immutable inside compute nodes (i.e., it is read only). Introducing a variable to Makefile to set the temp directory location might be better.

Thanks for your prompt reply. I added several candidates (`/tmp`, `/dev/shm`, and `pwd`). The function checks the existence, writable property, and noexec flag.

PR #297, opened 2024-05-09
feat: support assign gpu id for sanity test

cxz66666:
I find that the sanity test lack of `-d` param for assign GPU id, https://github.com/NVIDIA/gdrcopy/pull/112 already created a PR include add this feature, but that PR not progress for a long time.

So I create a new PR only include this specific feature.


pakmarkthub (approved):
Thank you. It is a nice-to-have option. Alternatively, we can set `CUDA_VISIBLE_DEVICES=<gpu_id> ./sanity` to achieve the same functionality.

PR #304, opened 2024-09-06
Add a new API to get DMA map of memory

ajitkhaparde:
Add a new API to get the DMA memory or the Physical Address of the peer mapped memory.
This will allow NIC devices to use the physical addresses directly instead of using the CUDA provided Unified Memory Address.

### DIFF
--- a/include/gdrapi.h
+++ b/include/gdrapi.h
@@ -79,6 +79,7 @@ typedef struct gdr_mh_s {
 // Create a peer-to-peer mapping of the device memory buffer, returning an opaque handle.
 // Note that at this point the mapping is still not accessible to user-space.
 int gdr_pin_buffer(gdr_t g, unsigned long addr, size_t size, uint64_t p2p_token, uint32_t va_space, gdr_mh_t *handle);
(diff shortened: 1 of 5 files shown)


pakmarkthub:
Hi @ajitkhaparde,

If my understanding is correct, `gdr_p2p_dma_map_buffer` is the same as `gdr_pin_buffer`. However, it also returns `paddr` of the cudaMalloc'd VA. Why do you need this information? Will you pass this `paddr` to your NIC stack?

At least for NVIDIA NICs, nvidia-peermem driver is used to provide this functionality. That PA and the corresponding DMA addresses are exchanged inside the kernel space. You can also use DMABUF instead of nvidia-peermem driver.

ajitkhaparde:
Yes. That is correct.
I am using that in the DPDK gpudev library.
At the end of it all, the PA is passed to the NIC driver for direct DMA to
the GPU memory.
Here is the patchwork link.
https://patches.dpdk.org/project/dpdk/list/?series=32920

Thanks
Ajit

drossetti:
FYI exposing phys addresses to user space is considered a security risk, hence this MR would not pass a security review. 
Depending on your use case, it might still be usable for you.

pakmarkthub:
Exposing PA to user space is considered a security risk. We cannot accept this change.
Closing it.

PR #321, opened 2025-03-31
Improve cuda gencode flags

agirault:
## Context

- The current cuda SASS/PTX list is hardcoded manually based on a versioning heuristic that is error-prone. Case in point:
  - blackwell sm 10.1 is missing
  - blackwell sm 10.0 is supported after 12.8, not 12.6. 
   ```make
   ```
- full arch list are used for code (SASS) and compute (PTX). For PTX, only latest is needed.

## Changes

- Use sm list from `nvcc --list-gpu-code` directly when available
- Fix blackwell sm list and version compatibility
- Consolidate compute & sm list in a single variable
- Only build PTX for last supported arch

### DIFF
--- a/scripts/get_cuda_gencode.sh
+++ b/scripts/get_cuda_gencode.sh
@@ -60,12 +60,14 @@ fi
 
 # Generate NVCC flags
 GENCODE_FLAGS=""
-for arch in $ARCH_LIST; do
-    GENCODE_FLAGS="$GENCODE_FLAGS -gencode arch=compute_$arch,code=compute_$arch"
-done
 
+# Generate SM-specific code for all architectures
 for arch in $ARCH_LIST; do
     GENCODE_FLAGS="$GENCODE_FLAGS -gencode arch=compute_$arch,code=sm_$arch"
 done
+
+# Generate PTX code only for the latest architecture
+LATEST_ARCH=$(echo "$ARCH_LIST" | tr ' ' '\n' | sort -n | tail -1)
+GENCODE_FLAGS="$GENCODE_FLAGS -gencode arch=compute_$LATEST_ARCH,code=compute_$LATEST_ARCH"
 
 echo "$GENCODE_FLAGS"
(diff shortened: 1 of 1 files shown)


scripts/get_cuda_gencode.sh
```diff
@@ -17,60 +17,57 @@ if [ "$CUDA_VERSION_MAJOR" -lt 8 ]; then
...
+    if [ "$CUDA_VERSION_MAJOR" -ge 12 ]; then
+        ARCH_LIST="$ARCH_LIST 90"
+    fi
 
-# Add Blackwell (12.0) if CUDA >= 12.8
-if [ "$CUDA_VERSION_MAJOR" -ge 12 ] && [ "$CUDA_VERSION_MINOR" -ge 8 ]; then
-    COMPUTE_LIST="$COMPUTE_LIST 120"
-    SM_LIST="$SM_LIST 120"
+    # Add Blackwell (10.0, 10.1, 12.0) if CUDA >= 12.8
+    if [ "$CUDA_VERSION_MAJOR" -ge 12 ] && [ "$CUDA_VERSION_MINOR" -ge 8 ]; then
```
drossetti, line 56: does this check work for CUDA 13.0?

scripts/get_cuda_gencode.sh
```diff
@@ -17,60 +17,57 @@ if [ "$CUDA_VERSION_MAJOR" -lt 8 ]; then
...
-    COMPUTE_LIST="$COMPUTE_LIST 100"
-    SM_LIST="$SM_LIST 100"
-fi
+    # Add Ampere (8.0, 8.6, 8.7) if CUDA >= 11.1
+    if [ "$CUDA_VERSION_MAJOR" -ge 11 ] && [ "$CUDA_VERSION_MINOR" -ge 1 ]; then
+        ARCH_LIST="$ARCH_LIST 80 86 87"
+    fi
+
+    # Add Ada Lovelace (8.9) if CUDA >= 11.8
+    if [ "$CUDA_VERSION_MAJOR" -ge 11 ] && [ "$CUDA_VERSION_MINOR" -ge 8 ]; then
```
drossetti, line 46: does this work on CUDA 12.0 ?

scripts/get_cuda_gencode.sh
agirault: same as above

pakmarkthub:
@agirault The fix has been merged to R2.5 branch (https://github.com/NVIDIA/gdrcopy/tree/R2.5)

PR #322, opened 2025-04-01
feat: return physical address base of memory range

ah-gus:
this patch, adds the ability to retrieve the PCIE space physical Address of a GPU allocated memory region.
it turns gdrcopy into the De-Facto standard way to implement GPUDirect RDMA to 3rd party PCIE cards, leveraging the good work done elsewhere in this project, the excellent bookkeeping and the NVIDIA sponsorship

### DIFF
--- a/include/gdrapi.h
+++ b/include/gdrapi.h
@@ -131,6 +131,7 @@ typedef struct gdr_info_v2 {
     unsigned mapped:1;
     unsigned wc_mapping:1;
     gdr_mapping_type_t mapping_type;
+    uint64_t paddr;
 } gdr_info_v2_t;
 typedef gdr_info_v2_t gdr_info_t;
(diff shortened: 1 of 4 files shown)


drossetti:
@ah-gus thanks for providing this PR.

> it turns gdrcopy into the De-Facto standard way to implement GPUDirect RDMA to 3rd party PCIE cards in userspace,

Although this approach may be deployed in specific environment, like embedded systems, it is not widely useful.

Exposing CPU physical addresses for the PCIe BAR of a device to user space a) has security implications and b) does not work on many platforms or in certain environments, for example in a VM or when the IOMMU is enabled.

Besides, the upstream Linux kernel solution for direct I/O on device memory is [dma-buf](https://www.kernel.org/doc/html/latest/driver-api/dma-buf.html). 3rd party devices should rather use those APIs.

As such I don't think it is wise to accept this change. It still makes sense to host that code on your repo and write a nice post somewhere.

